PR #2, opened 2022-09-08
CRYP-10: setup api

Andre-Ibrahim:
unit and e2e test where already setup with jest

eslint and prettier setup as minicap was

To Do: dockerize it


packages/api/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "nest-typescript-starter",
```
jason-gerard, line 2: Change this to be the package name so `@cryptify/api` this will make it so that yarn workspaces will recognize it as a package in the monorepo

packages/api/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
+    "test:e2e": "jest --config ./test/jest-e2e.json"
+  },
+  "dependencies": {
+    "@nestjs/common": "^9.0.0",
+    "@nestjs/core": "^9.0.0",
+    "@nestjs/platform-express": "^9.0.0",
+    "reflect-metadata": "^0.1.13",
+    "rimraf": "^3.0.2",
+    "rxjs": "^7.5.5"
```
jason-gerard, line 28: Is this package added by default?

packages/api/src/controllers/base/base.controller.ts
```diff
@@ -0,0 +1,12 @@
+import { Controller, Get } from "@nestjs/common";
+import { BaseService } from "../../services/base.service";
+
+@Controller("")
+export class BaseController {
+    constructor(private readonly baseService: BaseService) {}
+
+    @Get("")
+    ok(): string {
```
jason-gerard, line 9: Just so we are following the standard convention this GET method in a controller should be called `find()`

packages/api/src/services/base.service.ts
```diff
@@ -0,0 +1,8 @@
+import { Injectable } from "@nestjs/common";
+
+@Injectable()
+export class BaseService {
+    getOk(): string {
```
jason-gerard, line 5: The service would follow the same naming convention here too so `find()`

packages/api/tsconfig.build.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["node_modules", "dist", "test", "**/*spec.ts"]
```
jason-gerard, line 3: I'm not sure you need this file, I think you can just put the line `"exclude": ["node_modules", "dist", "test", "**/*spec.ts"]` in the tsconfig.json file after the "compilerOptions" key

jason-gerard (approved):

packages/api/package.json
Andre-Ibrahim: yup I didn't install any new dependencies

packages/api/tsconfig.build.json
Andre-Ibrahim: turns out this file is crucial for the app to run
image

jason-gerard: Ah ok I see, thanks for checking

packages/api/package.json
Andre-Ibrahim: Done!

PR #3, opened 2022-09-09
CRYP-9: Setup react native base with monorepo support

jason-gerard:
## Changes
- Add react native base project as an expo app
- Add yarn workspaces
- Integrate common package and api package in workspace


packages/client/.expo-shared/assets.json
```diff
@@ -0,0 +1,6 @@
+{
+  "e997a5256149a4b76e6bfd6cbf519c5e5a0f1d278a3d8fa1253022b03c90473b": true,
```
Andre-Ibrahim, line 2: do you know what this is used for?

Andre-Ibrahim (approved):

packages/client/.expo-shared/assets.json
jason-gerard: I believe it is a key value store of the optimized assets for expo. So it would include all the images, icons, etc.

PR #4, opened 2022-09-10
CRYP-9: Move eslint and prettier to project root

jason-gerard:
## Context
To make best use of the monorepo we can move all the general Typescript config which is project wide like formatting and linting into the root so all packages can take advantage

## Changes
- Move eslint and prettier files to root
- Add new scripts to project root
- Cleanup dependencies in packages


Andre-Ibrahim (approved):
makes sense

PR #5, opened 2022-09-10
CRYP-11: Setup docker and docker compose

jason-gerard:
## Changes
- Add docker files for
  - client web build
  - api
  - api database
- Add docker-compose file for dev environment


alexbara2000 (approved):
It is very good. However can we wait before creating database tables, there still needs to be a conversation about what we are storing and how.

jason-gerard:
> It is very good. However can we wait before creating database tables, there still needs to be a conversation about what we are storing and how.

No database tables are being made here, just a dockerfile for the database itself

packages/docker/api.Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM node:16-slim AS base
```
Andre-Ibrahim, line 1: 16-slim doesn't work for m1 should we make it 16 to avoid confusion

Andre-Ibrahim (approved):

packages/docker/api.Dockerfile
jason-gerard: The issue with that is only with the bcrypt library. But I think I found a solution for it we can try when we set it up for that user story.

Andre-Ibrahim: yea you're right, looking forward to seeing your solution!

PR #7, opened 2022-09-13
CRYP-21: Setup users db schema

jason-gerard:
## Changes
- Add entity for user using typeorm
- Refactor modules to allow for orm integration
- Implement config file and process injection

## Dev notes
- Updated dockerfile for api because the slim base image did not include some system level libraries required for hot reload on macos and linux (ubuntu)


packages/common/src/entities/user.ts
```diff
@@ -0,0 +1,22 @@
+import { Column, CreateDateColumn, Entity, PrimaryGeneratedColumn } from "typeorm";
+
+@Entity()
+export class User {
```
alexbara2000, line 4: So would this be 1 table in the database? Would we need to do the same concept for all the different entities that go in the db?

alexbara2000 (approved):

packages/common/src/entities/user.ts
jason-gerard: Yep and every entity we have in our system i.e. wallet, transaction, etc will have a file for them. Later on we will have many to many joins between our tables that will create a new join table without an explicit file.

packages/api/src/base/base.controller.ts
```diff
@@ -1,5 +1,5 @@
 import { Controller, Get } from "@nestjs/common";
-import { BaseService } from "../../services/base/base.service";
+import { BaseService } from "./base.service";
```
Andre-Ibrahim, line 2: why did you remove the folders for services and controllers?

Andre-Ibrahim (approved):

packages/api/src/base/base.controller.ts
jason-gerard: This is a more domain driven design approach of setting up the directories. Instead of splitting horizontally by layers we split vertically by resources / entities (the entity itself is stored in the shared common package though).

NestJS recommends it as well, if you don't set it up like this you have to do some weird stuff to get the repository code gen to work and I also wanted to try it out to see how we like it, if its ends up being worse its always an easy refactor.

Andre-Ibrahim: yea i found it odd that nest setup didn't have directories but your right should be a quick fix if we don't like it

Andre-Ibrahim: merge it
dew it!

PR #10, opened 2022-09-15
CRYP-49: Update yarn.lock from new deps

jason-gerard:


yarn.lock
```diff
@@ -17525,7 +17525,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"react-native-web@npm:^0.17.7":
```
alexbara2000, line 17528: What does this do?

alexbara2000 (approved):

yarn.lock
jason-gerard: If you prefix the version with a ^ it will pin the patch version only. Since the bug is in version 0.18.x we need to pin the minor version to 17 as well thats what the ~ does

PR #11, opened 2022-09-16
CRYP-48: implement api for create user

Andre-Ibrahim:
- added auth/sign in endpoint
- added test for controller method


packages/api/package.json
```diff
@@ -20,8 +20,15 @@
         "@nestjs/common": "^9.0.0",
         "@nestjs/config": "^2.2.0",
         "@nestjs/core": "^9.0.0",
+        "@nestjs/jwt": "^9.0.0",
+        "@nestjs/passport": "^9.0.0",
         "@nestjs/platform-express": "^9.0.0",
         "@nestjs/typeorm": "^9.0.1",
+        "bcrypt": "^5.0.1",
+        "class-transformer": "^0.5.1",
```
jason-gerard, line 28: Don't need this dep anymore
```suggestion
```

packages/api/src/authentication/authentication.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { AuthenticationController } from "./authentication.controller";
+import { AuthenticationService } from "@cryptify/api/src/authentication/authentication.service";
+import { UsersService } from "@cryptify/api/src/users/users.service";
+import { BadRequestException } from "@nestjs/common";
+
+describe("AuthenticationController::createUser", () => {
```
jason-gerard, line 7:
```suggestion
describe("AuthenticationController::signUp", () => {
```

packages/api/src/authentication/authentication.controller.ts
```diff
@@ -0,0 +1,28 @@
...
+import { signUpSchema } from "@cryptify/common/src/validations/user";
+import { AuthenticationService } from "@cryptify/api/src/authentication/authentication.service";
+import { AuthGuard } from "@nestjs/passport";
+
+@Controller("auth")
+export class AuthenticationController {
+    constructor(private readonly authService: AuthenticationService) {}
+
+    @Post("signup")
+    async signUp(@Body() user: User): Promise<any> {
```
jason-gerard, line 12: Add a type here

packages/api/src/authentication/authentication.module.ts
```diff
@@ -0,0 +1,27 @@
...
+import { JwtModule } from "@nestjs/jwt";
+import { JwtStrategy } from "@cryptify/api/src/authentication/jwt.strategy";
+import { ConfigModule } from "@nestjs/config";
+
+@Module({
+    controllers: [AuthenticationController],
+    imports: [
+        ConfigModule.forRoot({
+            envFilePath: `.env.${process.env.NODE_ENV}`,
+        }),
```
jason-gerard, line 16: I don't think you need this here, if it doesn't work without it set isGlobal: true in the app.module file where it is declared

packages/api/src/authentication/authentication.module.ts
```diff
@@ -0,0 +1,27 @@
...
+        UsersModule,
+        PassportModule.register({ defaultStrategy: "jwt", session: false }),
```
jason-gerard, line 18: Do you need to register a default strategy? will be cleaner if you don't need it

```suggestion
        PassportModule,
```

packages/api/src/authentication/authentication.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import { UsersService } from "../users/users.service";
+import { User } from "@cryptify/common/src/entities/user";
+import * as bcrypt from "bcrypt";
+import { JwtService } from "@nestjs/jwt";
+
+@Injectable()
+export class AuthenticationService {
+    constructor(private jwtService: JwtService, private usersService: UsersService) {}
+
+    async create(user: User): Promise<any> {
```
jason-gerard, line 11: I would call this reqUser and add a type to the return

```suggestion
    async create(reqUser: User): Promise<any> {
```

packages/api/src/authentication/authentication.service.ts
```diff
@@ -0,0 +1,30 @@
...
+        user.password = await bcrypt.hash(user.password, 10);
+
+        const createdUser = await this.usersService.create(user);
```
jason-gerard, line 14: Then call this user

```suggestion
        const user = await this.usersService.create(user);
```

packages/api/src/authentication/authentication.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+        const payload = { email: createdUser.email, sub: createdUser.id };
```
jason-gerard, line 16: I would just store the user id in the jwt payload unless there is a need for the email as well?

packages/api/src/base/base.module.ts
```diff
@@ -3,10 +3,11 @@ import { TypeOrmModule } from "@nestjs/typeorm";
 import { BaseController } from "@cryptify/api/src/base/base.controller";
 import { BaseService } from "@cryptify/api/src/base/base.service";
 import { User } from "@cryptify/common/src/entities/user";
+import { UsersService } from "@cryptify/api/src/users/users.service";
 
 @Module({
     imports: [TypeOrmModule.forFeature([User])],
-    providers: [BaseService],
+    providers: [BaseService, UsersService],
```
jason-gerard, line 10: Why is UserService added here?

packages/api/src/base/base.module.ts
```diff
@@ -3,10 +3,11 @@ import { TypeOrmModule } from "@nestjs/typeorm";
 import { BaseController } from "@cryptify/api/src/base/base.controller";
 import { BaseService } from "@cryptify/api/src/base/base.service";
 import { User } from "@cryptify/common/src/entities/user";
+import { UsersService } from "@cryptify/api/src/users/users.service";
 
 @Module({
     imports: [TypeOrmModule.forFeature([User])],
```
jason-gerard, line 9: Not sure we need this here anymore?

packages/client/package.json
```diff
@@ -20,6 +20,7 @@
         "@react-navigation/bottom-tabs": "^6.0.5",
         "@react-navigation/native": "^6.0.2",
         "@react-navigation/native-stack": "^6.1.0",
+        "class-validator": "^0.13.2",
```
jason-gerard, line 23: Not needed here
```suggestion
```

packages/api/src/authentication/authentication.controller.ts
```diff
@@ -0,0 +1,28 @@
...
+
+        if (error) {
+            throw new BadRequestException();
+        }
+
+        return await this.authService.create(user);
+    }
+
+    // To Do: re-write in sign in pr
+    @UseGuards(AuthGuard("local"))
```
jason-gerard, line 23: I'd make this its own Gaurd class like you did with the JwtGaurd, looks cleaner. May want to do that in the sign in pr though, up to you

packages/api/src/authentication/authentication.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    it("should return status 400 if user data is invalid", async () => {
+        user.email = "";
+        let error;
+        try {
+            await controller.signUp(user);
+        } catch (err) {
+            error = err;
+        }
+
+        expect(error).toBeInstanceOf(BadRequestException);
```
jason-gerard, line 61: A cleaner way to expect an error IMO

```suggestion
    expect(async () => {
        user.email = "";
        await controller.signUp(user);
    }).toThrow(BadRequestException);
```

packages/api/src/authentication/authentication.controller.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AuthenticationService } from "@cryptify/api/src/authentication/authentication.service";
+import { AuthGuard } from "@nestjs/passport";
+
+@Controller("auth")
+export class AuthenticationController {
+    constructor(private readonly authService: AuthenticationService) {}
+
+    @Post("signup")
+    async signUp(@Body() user: User): Promise<any> {
+        const { error } = signUpSchema.validate(user);
```
jason-gerard, line 13: I think this is best practice no?

```suggestion
        const { value, error } = signUpSchema.validate(user);
```

domsec:
Just wanted to chime in and let you know that sign up will no longer have a Remember Me checkbox. Therefore, the user should ideally remain signed in upon account creation for the length of the browser session (i.e. session instead of a cookie).

packages/api/src/authentication/authentication.controller.ts
Andre-Ibrahim: The signup returns a token, I'll make a new entity in common so that the client can use the same type

packages/api/src/authentication/authentication.module.ts
Andre-Ibrahim: already tried using isGlobal it doesn't work as intended the Secret becomes null will see if theres another field that needs to be set if not ill keep like that for now.

packages/api/src/authentication/authentication.module.ts
Andre-Ibrahim: was added when troubleshooting the JwtService issue, works without it!

packages/api/package.json
Andre-Ibrahim: you're right was using it when setting up NestJs validation instead of joi

packages/api/src/authentication/authentication.controller.ts
Andre-Ibrahim: added for both the controller and service method

packages/api/src/authentication/authentication.service.ts
Andre-Ibrahim: already added Token type from previous comment, and your naming makes more sense so ill update it my code.

packages/api/src/authentication/authentication.service.ts
Andre-Ibrahim: This is just following the NestJs documentation, i guess we don't need it but its a nice to have in case we need access we wouldn't need a database lookup, unless you think it adds to much overhead

packages/api/src/base/base.module.ts
Andre-Ibrahim: I double checked and the UsersService is not used in the scope of the base module so Will be removed.

packages/api/src/base/base.module.ts
Andre-Ibrahim: that import was used to inject the UsersService as a provider but since we no longer need it it can be removed, smae for the TypeOrmModule since the usersRepository isn't being used (and the User entity).

packages/client/package.json
Andre-Ibrahim: removed!

packages/api/src/authentication/authentication.controller.ts
Andre-Ibrahim: yea will make that changes in the signIn pr, it can be useful if we want to add our own logic

packages/api/src/authentication/authentication.controller.spec.ts
Andre-Ibrahim: I tried to do it like that at first but the test fails am i missing something here?

```ts
 it("should return status 400 if user data is invalid", async () => {
        expect(async () => {
            user.email = "";
            await controller.signUp(user);
        }).toThrow(BadRequestException);
    });
```

packages/api/src/authentication/authentication.controller.ts
Andre-Ibrahim: right was using the user var to create the new user through the authService but it makes more sense to use "value"

Andre-Ibrahim:
> Just wanted to chime in and let you know that sign up will no longer have a Remember Me checkbox. Therefore, the user should ideally remain signed in upon account creation for the length of the browser session (i.e. session instead of a cookie).

That would be handled in the front end, our api will return an access token that will be valid for a certain amount of time(currently set to 2hrs but could be modified) and the client can handle that token however it wants.

packages/api/src/authentication/authentication.service.ts
jason-gerard: Check out the RFC for JWTs https://www.rfc-editor.org/rfc/rfc7519#page-8

Basically here is the list of "claims" that can be made by a JWT, none of these claims are mandatory for the token to have.
- iss (issuer): identifies the principal that issued the JWT.
- sub (subject): identifies the principal that is the subject of the JWT. Must be unique
- aud (audience): identifies the recipients that the JWT is intended for (array of strings/uri)
- exp (expiration time): identifies the expiration time (UTC Unix) after which you must no longer accept this token. It should be after the issued-at time.
- nbf(not before): identifies the UTC Unix time before which the JWT must not be accepted
- iat (issued at): identifies the UTC Unix time at which the JWT was issued
- jti (JWT ID): provides a unique identifier for the JWT.

So we could have this to start, and add in other parts as we need them
```
{
    "iss": "cryptify",
    "sub": user.id,
    "context": {
        "user": {
            "id": user.id,
            "email": user.email
        },
        "roles":["basic"]
    }
}
```

packages/api/src/authentication/authentication.controller.spec.ts
jason-gerard: Sorry the above snippet was for sync methods, for testing async methods that throws errors you can do it like below where you manually reject the promise.

```js
it("should return status 400 if user data is invalid", async () => {
    user.email = "";
    await expect(controller.signUp(user))
        .rejects
        .toThrow(BadRequestException);
});

Comment (Andre-Ibrahim) [reply packages/api/src/authentication/authentication.controller.spec.ts:61]: right it works now!

File: packages/api/src/authentication/authentication.module.ts [hunk 15]
```diff
@@ -0,0 +1,27 @@
...
+import { UsersModule } from "@cryptify/api/src/users/users.module";
+import { PassportModule } from "@nestjs/passport";
+import { JwtModule } from "@nestjs/jwt";
+import { JwtStrategy } from "@cryptify/api/src/authentication/jwt.strategy";
+import { ConfigModule } from "@nestjs/config";
+
+@Module({
+    controllers: [AuthenticationController],
+    imports: [
+        ConfigModule.forRoot({
```
jason-gerard, line 14: just a reminder for myself to look into this

jason-gerard (approved):

packages/api/src/authentication/authentication.module.ts
Andre-Ibrahim: sure, ill have a second look at it too in the next pr

domsec:
> > Just wanted to chime in and let you know that sign up will no longer have a Remember Me checkbox. Therefore, the user should ideally remain signed in upon account creation for the length of the browser session (i.e. session instead of a cookie).
> 
> That would be handled in the front end, our api will return an access token that will be valid for a certain amount of time(currently set to 9999 years but could be modified) and the client can handle that token however it wants.

Alrighty good with me.

PR #12, opened 2022-09-16
CRYP-56: create basic ERD of the system

alexbara2000:
Added the basic ERD


Andre-Ibrahim (commented):
are we differentiating the "pros" in our system?
Also if we are try to do a 1 to 1 mapping with our database we have a created_at timestamp field for the user entity.

https://github.com/SOEN-490-Capstone/Cryptify/blob/main/packages/common/src/entities/user.ts

alexbara2000:
@Andre-Ibrahim There is a bool in the user called "is_pro" which is how we will differentitate the pro users. For the created_at field, I just added it. I don't think it is fully necessary since this is high level and a lot more fields are going to be added in the future, having consistency between the entities and the database from early on is a good idea.

Andre-Ibrahim:
yes but the "pro" user is something we described in the user persona to identify how the user will interact with our system I don't think we are planing to make different data/pages available based on that so I see no use case for it.

Andre-Ibrahim (approved):
looks good

domsec:
@Andre-Ibrahim My understanding is we will in fact be displaying certain data/pages based on whether the user is basic or pro.

Andre-Ibrahim:
yes @domsec  just looked at our backlog deleted my comment before you refresh lol

packages/diagrams/erd.md
```diff
@@ -1,19 +1,26 @@
 ```mermaid
 erDiagram
-    CUSTOMER ||--o{ ORDER : places
-    CUSTOMER {
+    USERS ||--o{ WALLETS : has
+    USERS {
+        int user_id
         string name
```
domsec, line 6: Name should be split into first name and last name since first name will be used in the navigation bar.

PR #13, opened 2022-09-18
Cryp 22 as a user i want to be able to sign in backend

alexbara2000:
Added ability to sign in as a user. The api will return a jwt token if the password was correct.

### DIFF
--- a/packages/api/src/authentication/authentication.controller.ts
+++ b/packages/api/src/authentication/authentication.controller.ts
@@ -21,9 +21,12 @@ export class AuthenticationController {
     }
 
     // To Do: re-write in sign in pr
-    @UseGuards(AuthGuard("local"))
     @Post("signin")
-    async signIn(@Req() req): Promise<User> {
-        return req.user;
+    async signIn(@Body() email: string, @Body() password:  string): Promise<User> {
+        const val = this.authService.validateUser(email, password);
+        if (val == null) {
+            throw new BadRequestException();
+        }
+        return val;
     }
 }
--- a/packages/api/src/authentication/authentication.controller.ts
+++ b/packages/api/src/authentication/authentication.controller.ts
@@ -4,6 +4,7 @@ import { signUpSchema } from "@cryptify/common/src/validations/user";
 import { AuthenticationService } from "@cryptify/api/src/authentication/authentication.service";
 import { AuthGuard } from "@nestjs/passport";
 import { Token } from "@cryptify/common/src/types/token";
+import { Console } from "console";
 
 @Controller("auth")
 export class AuthenticationController {
@@ -20,13 +21,9 @@ export class AuthenticationController {
         return await this.authService.create(value);
     }
 
-    // To Do: re-write in sign in pr
     @Post("signin")
-    async signIn(@Body() email: string, @Body() password:  string): Promise<User> {
-        const val = this.authService.validateUser(email, password);
-        if (val == null) {
-            throw new BadRequestException();
-        }
-        return val;
+    @UseGuards(AuthGuard("local"))
+    async signIn(@Req() req): Promise<Token> {
+        return this.authService.validateUser(req.body.email, req.body.password);
     }
 }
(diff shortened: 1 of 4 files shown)


packages/api/src/authentication/authentication.controller.spec.ts
```diff
@@ -43,14 +48,30 @@ describe("AuthenticationController::signUp", () => {
...
+    it("Should return an access token if the user can sign in",  async () => {
+        //wont work cuase we expcet a req not a user
+        expect(await controller.signIn(userSignIn)).toStrictEqual(result);
+    });
+
+    it("Should returnstatus 400 if user can't sign in",  async () => {
+        //wont work cuase we expcet a req not a user
+        user.password = "wrong_password";
+        expect(await controller.signIn(userSignIn)).toStrictEqual({ access_token: "token" });
+    });
```
jason-gerard, line 76: These should be looked at again, not sure this is testing properly

jason-gerard: These should also be moved to a separate "describes" block for `AuthenticationController::signIn`

packages/api/src/authentication/authentication.controller.ts
```diff
@@ -20,10 +21,14 @@ export class AuthenticationController {
         return await this.authService.create(value);
     }
 
-    // To Do: re-write in sign in pr
-    @UseGuards(AuthGuard("local"))
     @Post("signin")
-    async signIn(@Req() req): Promise<User> {
-        return req.user;
+    @UseGuards(AuthGuard("local"))
```
jason-gerard, line 25: Move this guard decorator above the @Post decorator so its standardized

packages/api/src/authentication/authentication.controller.ts
```diff
@@ -20,10 +21,14 @@ export class AuthenticationController {
...
+    async signIn(@Body() user: UserCredential): Promise<Token> {
+        const { value, error } = signInSchema.validate(user);
+
+        if (error) {
+            throw new BadRequestException();
+        }
+        return this.authService.validateUser(user.email, user.password);
```
jason-gerard, line 32: Use the value object not the body object because its the processed value for convention

```suggestion
        return this.authService.validateUser(value.email, value.password);
```

packages/common/src/types/UserCredential.ts
```diff
@@ -0,0 +1,4 @@
+export type UserCredential = {
```
jason-gerard, line 1: Move this file to a new folder called `requests` and call it the request method name + "Request"

```suggestion
export type SignInRequest = {
```

packages/api/src/authentication/authentication.service.ts
```diff
@@ -19,12 +19,11 @@ export class AuthenticationService {
...
     }
 
-    // To Do: clean in sign in pr
-    async validateUser(email: string, pass: string): Promise<any> {
+    async validateUser(email: string, password: string): Promise<Token> {
         const user = await this.usersService.findOne(email);
-        if (user && user.password === pass) {
-            const { ...result } = user;
-            return result;
+        if (user && await bcrypt.compare(password, user.password)) {
```
jason-gerard, line 24:
```suggestion
        if (await bcrypt.compare(password, user?.password)) {
```

packages/diagrams/erd.md
```diff
@@ -3,7 +3,8 @@ erDiagram
     USERS ||--o{ WALLETS : has
     USERS {
         int user_id
-        string name
+        string first_name
```
jason-gerard, line 6: There should be another markdown file included here in the generated folder, make sure to add it

jason-gerard: Nevermind I see it was added before

packages/api/src/authentication/authentication.controller.spec.ts
```diff
@@ -46,11 +46,60 @@ describe("AuthenticationController::signUp", () => {
     });
 
     it("should return an access token if user data is valid", async () => {
-        expect(await controller.signUp(user)).toStrictEqual({ access_token: "token" });
+        expect(await controller.signUp(user)).toStrictEqual(result);
```
jason-gerard, line 49: This should be renamed to `accessToken` to keep the casing

jason-gerard (approved):

PR #14, opened 2022-09-19
CRYP-48: added type to sign up endpoint

Andre-Ibrahim:
- added SignUpRequest type to common
- move the test files in a test directory
- refactored the authentication controller test


packages/common/src/types/requests/signUpRequest.ts
```diff
@@ -0,0 +1,6 @@
+export type SignUpRequest = {
```
jason-gerard, line 1: File names are with underscores so sign_up_request.ts 

Also need to update the other request type file.

jason-gerard (approved):

alexbara2000 (approved):
Looks good

PR #16, opened 2022-09-21
CRYP-21: Migrate and refactor to yup from joi

jason-gerard:
## Changes
- Migrate joi package to yup
- Create custom validator hook
- Use type gen from yup instead of re-creating types


packages/common/src/validations/sign_up_schema.ts
```diff
@@ -0,0 +1,11 @@
+import * as yup from "yup";
+import { passwordSchema } from "@cryptify/common/src/validations/password_schema";
+
+export const signUpSchema = yup
+    .object({
+        firstName: yup.string().matches(new RegExp("^[A-Za-z]+")).min(0).max(32).required(),
+        lastName: yup.string().matches(new RegExp("^[A-Za-z]+")).min(0).max(32).required(),
+        email: yup.string().email().required(),
+        password: passwordSchema,
```
narroarrow, line 9: is confirmPassword needed here?

jason-gerard: Should be added in your pull request since it's something new for the frontend.

Andre-Ibrahim (approved):
Looks good

PR #18, opened 2022-09-23
CRYP-21: Fix mobile app integration

jason-gerard:
## Changes
- Remove named volume for postgres with docker compose
- Add expo entry point for app
- Add crypto math pollyfill


packages/client/gateways/request.ts
```diff
@@ -1,5 +1,5 @@
 export async function request<T>(path: string, req: any): Promise<T> {
-    const response = await fetch(`http://localhost:3001/${path}`, {
+    const response = await fetch(`http://192.168.86.35:3001/${path}`, {
```
Andre-Ibrahim, line 2: we should probably move the ip to an env file later

packages/client/index.js
```diff
@@ -1,3 +1,4 @@
+import "react-native-get-random-values";
```
Andre-Ibrahim, line 1: is that being used anywhere?

Andre-Ibrahim (approved):
looks good

packages/client/index.js
jason-gerard: This is the pollyfill that will replace the crypto math function that isn't supported by react native, it is called there so when other internal library functions call the math functions it will be overridden already.

packages/client/gateways/request.ts
jason-gerard: Yep this will be different per person, will update this before merging

PR #20, opened 2022-09-25
CRYP-51: Fix sign in page styling to match UI mockups

domsec:
- Created custom theme files for each component used to match the UI mockups
- Refactored the sign in page to take into account these new additions

image
image

### DIFF
--- a/packages/client/theme/components/button.ts
+++ b/packages/client/theme/components/button.ts
@@ -1,0 +1,32 @@
+export const button = {
+    baseStyle: {
+        borderRadius: '100',
+        height: '50',
+        _text: {
+            fontWeight: 'semibold',
+        },
+        _loading: {
+            opacity: '60',
+        },
+    },
+    variants: {
+        solid: {
+            bg: 'darkBlue.500',
+            _text: {
+                color: 'white',
+            },
+            _hover: {
+                bg: 'darkBlue.500',
+                opacity: '60',
+            },
+            _pressed: {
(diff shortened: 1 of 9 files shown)


packages/common/src/validations/password_schema.ts
```diff
@@ -4,8 +4,10 @@ import * as yup from "yup";
 // has a minimum length of 8 characters
 export const passwordSchema = yup
     .string()
+    .min(6)
```
jason-gerard, line 7: Do you not need to add there error message here as well?
```suggestion
    .min(6, "Password must be between 6 and 20 characters.")
```

packages/client/theme/components/button.ts
```diff
@@ -0,0 +1,32 @@
+export const button = {
+    baseStyle: {
+        borderRadius: '100',
```
jason-gerard, line 3: Can you run `yarn run lint:fix` in the root of the project. This will fix all the code formatting and linting stuff like new line at eof, double quotes, etc

jason-gerard (approved):

packages/client/theme/components/button.ts
domsec: done

packages/common/src/validations/password_schema.ts
domsec: Yup, you're right. It's fixed

jason-gerard (approved):

PR #24, opened 2022-09-25
CRYP-21: Add API error messages to sign up and sign in

jason-gerard:
## Changes
- Add API error message for sign up and sign in
- Fix safe area and line height on sign up and sign in top text
- Refactor error messages to common
- Refactor request method to allow for different methods

## Screenshots
image
image

### DIFF
--- a/packages/client/screens/SignInScreen.tsx
+++ b/packages/client/screens/SignInScreen.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { StyleSheet } from "react-native";
-import { Input, Button, VStack, FormControl, Pressable, Text } from "native-base";
+import { Input, Button, VStack, FormControl, Pressable, Text, Box } from "native-base";
 import { View } from "../components/Themed";
 import { Formik } from "formik";
 import { FontAwesomeIcon } from "@fortawesome/react-native-fontawesome";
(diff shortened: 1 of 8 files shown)


packages/client/screens/SignInScreen.tsx
```diff
@@ -76,6 +85,7 @@ export default function SignInScreen({ navigation }: RootTabScreenProps<"SignInS
 const styles = StyleSheet.create({
     title: {
         fontSize: 28,
+        lineHeight: 32,
```
domsec, line 88: I looked into the default Text component theme (https://github.com/GeekyAnts/NativeBase/blob/master/src/theme/components/text.ts). By default line height is set to 1.5rem (24px) so unless we cannot set it to auto (i.e. `line-height: normal;`) within our custom Text component theme, leave this as is. Also, probably best to just set line height to auto if the text no longer cuts cut off.

packages/common/src/errors/error_messages.ts
```diff
@@ -0,0 +1,3 @@
+export const ERROR_EMAIL_OR_PASSWORD_INCORRECT = "Your email or password was incorrect.";
```
domsec, line 1: Great idea to consolidate the error messages in a single file. We should probably do this for all error messages (i.e. sign in/sign up validation, etc.)

domsec (approved):

packages/client/screens/SignInScreen.tsx
jason-gerard: So its recommended that the line height be around 4-5px larger than whatever the text size is, since we can't dynamically set that in the theme, as far as I'm aware I think we might have to leave it here. We could always also just extract this out to a CustomTopText.tsx component and do it in there.

Andre-Ibrahim (approved):

PR #26, opened 2022-09-26
CRYP-72: Added eth-edge infrastructure

alexbara2000:
Added eth-edge infrastructure and all dependencies


packages/eth-edge/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
```
jason-gerard, line 1: Delete this file, this is all handled in the root of the project

packages/eth-edge/.gitignore
```diff
@@ -0,0 +1,35 @@
+# compiled output
```
jason-gerard, line 1: This file can be deleted too snce there is a global gitignore in the project root

packages/eth-edge/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
jason-gerard, line 1: This file can also be deleted since it is managed in the project root

packages/eth-edge/README.md
```diff
@@ -0,0 +1,73 @@
+<p align="center">
```
jason-gerard, line 1: This file can also be deleted

packages/eth-edge/src/main.ts
```diff
@@ -0,0 +1,8 @@
+import { NestFactory } from '@nestjs/core';
+import { AppModule } from './app.module';
+
+async function bootstrap() {
+  const app = await NestFactory.create(AppModule);
+  await app.listen(80);
```
jason-gerard, line 6: This should be done using an environment variable like in the api package but it can be added later

packages/eth-edge/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "baseUrl": "./",
+    "incremental": true,
+    "skipLibCheck": true,
+    "strictNullChecks": false,
+    "noImplicitAny": false,
+    "strictBindCallApply": false,
+    "forceConsistentCasingInFileNames": false,
+    "noFallthroughCasesInSwitch": false
+  }
+}
```
jason-gerard, line 21: This tsconfig should extend the root ts config and then apply its options, below is a good base

```suggestion
{
    "extends": "../../tsconfig.json",
    "compilerOptions": {
        "module": "commonjs",
        "target": "es2017",
        "declaration": true,
        "removeComments": true,
        "emitDecoratorMetadata": true,
        "experimentalDecorators": true,
        "allowSyntheticDefaultImports": true,
        "esModuleInterop": true,
        "incremental": true,
        "skipLibCheck": true,
        "sourceMap": true,
        "outDir": "./dist"
    },
    "include": ["./src"]
}
```

packages/eth-edge/src/app.controller.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { Test, TestingModule } from '@nestjs/testing';
```
jason-gerard, line 1: The unit and integration tests should be moved to mirror the same folder structure that is in the api packages

/test/
  - /unit/
  - /integration/

jason-gerard: Some other files and configuration for the paths will need to be updated for this as well, it is already done in the api package so you can just use that as reference

jason-gerard: Note: can be done later in another PR

packages/eth-edge/src/main.ts
jason-gerard: Note: can be done later in another PR

jason-gerard (approved):

PR #27, opened 2022-09-27
CRYP-18: Setup backend integration testing

jason-gerard:
## Changes
- Update docker to use new test-db for running tests with a live db
- Setup db restore scripts to reset the db before each test run
- Write integration tests for sign up and sign in endpoints
- Write documentation on how to run backend unit and integration tests, docs on frontend unit + snapshot tests and system tests to come

## Screenshots
image


Andre-Ibrahim (approved):

packages/api/test/integration/db_helpers/clear_db.ts
```diff
@@ -0,0 +1,12 @@
+import { DataSource } from "typeorm";
```
Andre-Ibrahim, line 1: should db_helpers be in common since the db is shared?

Andre-Ibrahim (approved):
just one question rest looks good to me!

packages/api/test/integration/db_helpers/clear_db.ts
jason-gerard: Yep will move it since tbh I think we should have integration tests in the edge services as well

PR #30, opened 2022-09-28
CRYP-58: create db schema for wallets

Andre-Ibrahim:
- added wallet entity in common


jason-gerard (commented):
Try to make some users and wallets in a test file then query them, see if it all works and that the resulting db schema is correct. Typeorm many times only gives errors during runtime because it all happens in the decorators

packages/common/src/entities/wallet.ts
```diff
@@ -0,0 +1,15 @@
+import { Entity, JoinColumn, ManyToOne, PrimaryColumn } from "typeorm";
+import { User } from "@cryptify/common/src/entities/user";
+
+@Entity()
+export class Wallet {
+    @PrimaryColumn()
+    walletAddress: string;
```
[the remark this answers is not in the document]
jason-gerard: TBH I'm not sure about this maybe address is better, wdyt?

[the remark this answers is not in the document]
Andre-Ibrahim: Meta mask call it address
image

Andre-Ibrahim:
> Try to make some users and wallets in a test file then query them, see if it all works and that the resulting db schema is correct. Typeorm many times only gives errors during runtime because it all happens in the decorators

is that something you think should be  included in the repo or just validation on my side?

jason-gerard:
> > Try to make some users and wallets in a test file then query them, see if it all works and that the resulting db schema is correct. Typeorm many times only gives errors during runtime because it all happens in the decorators
> 
> is that something you think should be included in the repo or just validation on my side?

Doesn't need to be checked in just so we are certain that typeorm is doing what we think its doing

packages/common/src/entities/wallet.ts
```diff
@@ -0,0 +1,15 @@
+import { Entity, JoinColumn, ManyToOne, PrimaryColumn } from "typeorm";
+import { User } from "@cryptify/common/src/entities/user";
+
+@Entity()
+export class Wallet {
```
jason-gerard, line 5: Should add another field here for the wallet name, not sure what a good word for it is, maybe just "name" makes sense

packages/common/src/entities/wallet.ts
```diff
@@ -0,0 +1,18 @@
+import { Column, Entity, JoinColumn, ManyToOne, PrimaryColumn } from "typeorm";
+import { User } from "@cryptify/common/src/entities/user";
+
+@Entity()
+export class Wallet {
+    @PrimaryColumn()
+    walletAddress: string;
```
jason-gerard, line 7: For here and for the `walletName` should we just drop the wallet prefix since it is already in the wallet entity? Seems kinda redundant

jason-gerard (approved):

packages/common/src/entities/wallet.ts
domsec: I agree

domsec (approved):

PR #31, opened 2022-09-29
CRYP-13: setup CI/CD pipeline

jason-gerard:
## Context
This pr adds a CI pipeline through github actions, the pipeline covers running build, lint, unit tests, integration tests, etc

## Changes
- Add CI pipeline `.yml` file
- Refactor configs so the `env` vars are not being used directly from the `process`


domsec (approved):
Looks good

packages/api/src/authentication/authentication.module.ts
```diff
@@ -5,19 +5,19 @@ import { UsersModule } from "../users/users.module";
 import { PassportModule } from "@nestjs/passport";
 import { JwtModule } from "@nestjs/jwt";
 import { JwtStrategy } from "./strategies/jwt.strategy";
-import { ConfigModule } from "@nestjs/config";
+import { ConfigService } from "@nestjs/config";
 
 @Module({
     controllers: [AuthenticationController],
     imports: [
-        ConfigModule.forRoot({
```
Andre-Ibrahim, line 13: can you validate the the env variables are read correctly, when i tried to set isGlobal to true in the app.module before i was having issues with that.

Andre-Ibrahim (approved):

packages/api/src/authentication/authentication.module.ts
jason-gerard: Yea basically since the config module is loaded async if you want to set isGlobal for config you need to register all the modules that use it as async with a dep on config (see JwtModule.registerAsync or TypeOrmModule.forRootAsync).

That allows us to inject the config service properly as well.

And you can see the only place we use the process.env directly now is in the db helpers where we don't have access to the container through the template method `dataSourceOptionsTemplateMethod`

jason-gerard: Also the integration tests passing in the CI pipeline kinda verify that for us

PR #32, opened 2022-09-29
CRYP-95: add basic settings page and navigation

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-95

## Changes
- Add settings screen
- Add screen to navigation
- Add header styling for blank page

## Screenshots
image

### DIFF
--- a/packages/client/navigation/index.tsx
+++ b/packages/client/navigation/index.tsx
@@ -17,6 +17,7 @@ import SignUpScreen from "../screens/SignUpScreen";
 import { RootStackParamList, RootTabParamList } from "../types";
 import LinkingConfiguration from "./LinkingConfiguration";
 import SignInScreen from "../screens/SignInScreen";
+import SettingsScreen from "../screens/SettingsScreen";
 
 export default function Navigation({ colorScheme }: { colorScheme: ColorSchemeName }) {
     return (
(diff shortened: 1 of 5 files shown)


Andre-Ibrahim (approved):
lgtm

domsec:
There shouldn't be a grey line present.

packages/client/navigation/index.tsx
```diff
@@ -65,6 +66,20 @@ function BottomTabNavigator() {
                     tabBarIcon: ({ color }) => <TabBarIcon name="code" color={color} />,
                 }}
             />
+            <BottomTab.Screen
+                name="SettingsScreen"
+                component={SettingsScreen}
+                options={{
+                    title: "Settings",
```
domsec, line 73: Is it possible to use the Text component in order to take advantage of the custom theme?

jason-gerard:
> There shouldn't be a grey line present.

I can't even see it lol, i'll try to change the drop shadow of the header i think it might work.

packages/client/navigation/index.tsx
```diff
@@ -65,6 +66,20 @@ function BottomTabNavigator() {
                     tabBarIcon: ({ color }) => <TabBarIcon name="code" color={color} />,
```
domsec, line 66: should be a menu icon

jason-gerard: Yea this should all get taken care of in the user story for the navbar. ideally he will make some component for future use with all the styling working.

packages/client/navigation/index.tsx
jason-gerard: Im reading now, it looks like there should be a way, let me test it out in a bit

packages/client/navigation/index.tsx
domsec: Perfect.

packages/client/navigation/index.tsx
jason-gerard: Ok so basically would have to make a completely new Text component anyways so not sure if its worth it

PR #33, opened 2022-10-01
CRYP-24: create sequence diagram for adding a new wallet

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-24


domsec (commented):
Success screen is only displayed if all wallet info (balance, transactions, etc.) is obtained and stored in the db. Also error screen sequence is missing.

jason-gerard:
> Success screen is only displayed if all wallet info (balance, transactions, etc.) is obtained and stored in the db. Also error screen sequence is missing.

The other parts can take a while to process so it doesn't really make sense to block the UI on them, also if they were to error that would mean there was some critical error.

I'll add the error screen sequence on this one, another separate diagram can be made to highlight some specific error for some other part of the flow but it is not really needed, just don't want to make one diagram too large because it is hard to read. In general sequence diagrams should show the main success flow and only include expected errors i.e. no point to take into account that a service is unreachable for example.

Andre-Ibrahim (approved):

PR #34, opened 2022-10-02
CRYP-59: backend api for adding a wallet

Andre-Ibrahim:


packages/common/src/entities/wallet.ts
```diff
@@ -12,6 +13,9 @@ export class Wallet {
     @Column()
     name: string;
 
+    @Column({ type: "enum", enum: CurrencyTypeEnum, default: CurrencyTypeEnum.Eth })
```
jason-gerard, line 16: Default here doesn't really make sense, one should just always be passwed

```suggestion
    @Column({ type: "enum", enum: CurrencyTypeEnum })
```

packages/common/src/enums/currency_type_enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum CurrencyTypeEnum {
+    Eth = "ETH",
+    Btc = "BTC",
+}
```
jason-gerard, line 4:
```suggestion
export enum CurrencyType {
    ETHEREUM = "ETHEREUM",
    BITCOIN = "BITCOIN",
}
```

packages/common/src/validations/create_wallet_schema.ts
```diff
@@ -0,0 +1,9 @@
+import * as yup from "yup";
+
+export const createWalletSchema = yup
+    .object({
+        address: yup.string().required("Enter a valid wallet address."),
+        name: yup.string().required("Enter a valid wallet name."),
+        currencyType: yup.string(),
```
jason-gerard, line 7: Should be able to make this verify it is one of the enum type

packages/api/test/unit/wallets/wallets.controller.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getRepositoryToken } from "@nestjs/typeorm";
+import { Wallet } from "@cryptify/common/src/entities/wallet";
+import { InsertResult, Repository } from "typeorm";
+import { CreateWalletRequest } from "@cryptify/common/src/requests/create_wallet_request";
+
+describe("WalletsController", () => {
+    let controller: WalletsController;
+    let fakeWalletsService: Partial<WalletsService>;
+    let fakeWalletsRepository: Partial<Repository<Wallet>>;
+    const result: InsertResult = { identifiers: [{ id: 1 }], generatedMaps: [], raw: [] };
```
jason-gerard, line 13: Not sure what is going on here

packages/api/test/integration/wallets_api.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    it("POST /wallets", async () => {
+        const loginRes = await agent(app.getHttpServer()).post("/auth/signup").send({
+            firstName: "fname",
+            lastName: "lname",
+            email: "test@test.com",
+            password: "Test123!",
+            confirmPassword: "Test123!",
+        });
+
+        const cookies = loginRes.header("set-cookie");
```
jason-gerard, line 32: Why set cookie? The token is just returned in the body

packages/api/src/wallets/wallets.service.ts
```diff
@@ -10,7 +12,13 @@ export class WalletsService {
         private walletRepository: Repository<Wallet>,
     ) {}
 
-    async create(userId: number, address: string, name: string): Promise<InsertResult> {
-        return this.walletRepository.insert({ userId, address, name });
+    async create(userId: number, createWalletReq: CreateWalletRequest): Promise<InsertResult> {
```
jason-gerard, line 15: Return type should be the wallet

packages/api/src/authentication/authentication.service.ts
```diff
@@ -14,9 +13,8 @@ export class AuthenticationService {
 
     async signUp(signUpReq: SignUpRequest): Promise<JwtToken> {
         signUpReq.password = await bcrypt.hash(signUpReq.password, 10);
-        const user = await this.usersService.create(signUpReq);
-
-        return this.signToken(user);
+        const userId = (await this.usersService.create(signUpReq)).identifiers[0].id;
```
jason-gerard, line 16: ???

packages/api/src/authentication/authentication.service.ts
```diff
@@ -30,11 +28,11 @@ export class AuthenticationService {
             throw new ForbiddenException(ERROR_EMAIL_OR_PASSWORD_INCORRECT);
         }
 
-        return this.signToken(user);
+        return this.signToken(user.id);
     }
 
-    private signToken(user: User): JwtToken {
-        const payload = { sub: user.id };
+    private signToken(userId: number): JwtToken {
```
jason-gerard, line 34: I'd leave this as the type User, in the future if we want to sign the token with other info about the user it will be available

packages/api/src/wallets/wallets.controller.ts
```diff
@@ -1,14 +1,20 @@
...
 
 @Controller("wallets")
 export class WalletsController {
     constructor(private walletsService: WalletsService) {}
 
     @Post()
-    async create(@Body() body: Wallet): Promise<InsertResult> {
-        return this.walletsService.create(body.userId, body.address, body.name);
+    @UseGuards(JwtAuthGuard)
+    async create(@Body() body: CreateWalletRequest, @Request() req): Promise<InsertResult> {
```
jason-gerard, line 15: POST /user/:id/wallet

packages/api/src/wallets/helpers/whatCryptoCurrencyType.ts
```diff
@@ -0,0 +1,7 @@
+import { CurrencyTypeEnum } from "@cryptify/common/src/enums/currency_type_enum";
+
+export function whatCryptoCurrencyType(walletAddress: string): CurrencyTypeEnum {
```
jason-gerard, line 3: This should probably also be in the common package

```suggestion
export function getCurrencyType(walletAddress: string): CurrencyTypeEnum {
```

packages/api/src/wallets/wallets.service.ts
```diff
@@ -10,7 +12,13 @@ export class WalletsService {
...
+        const createdWallet = this.walletRepository.create({
+            ...createWalletReq,
+            currencyType: whatCryptoCurrencyType(createWalletReq.address),
```
jason-gerard, line 18: type `CreateWalletRequest` contains the currencyType, don't think you need to re-calculate it

packages/common/src/validations/create_wallet_schema.ts
jason-gerard, line 7: This should also have the userId

packages/api/src/wallets/wallets.service.ts
```diff
@@ -10,7 +12,13 @@ export class WalletsService {
...
+            userId,
```
jason-gerard, line 19: The user id should be used from the request obj, just add a step before this in the controller probably in a hook to verify that the req user id is the same as the one in the token

jason-gerard (commented):
I see you are trying to make 1 pr per subtask, its fine if you think it will be too large on its own but its a bit weird because half of this code will get deleted / moved when you do it in eth edge

packages/api/test/unit/wallets/wallets.controller.spec.ts
Andre-Ibrahim: > I see you are trying to make 1 pr per subtask, its fine if you think it will be too large on its own but its a bit weird because half of this code will get deleted / moved when you do it in eth edge

do you think creating the wallet entity in the db should be done in eth-edge?
I thought it was the api's responsibility since it is not crypto specific.
Will refactor the code based on your sequence diagram

packages/api/test/integration/wallets_api.spec.ts
Andre-Ibrahim: was troubleshooting this doesn't even work lol but for some reason nest hides the accessToken at some point

packages/api/test/unit/wallets/wallets.controller.spec.ts
jason-gerard: Since the wallet is technically an ethereum wallet it should be handled in EthEdge, also since the actual "wallet creation" process might be very different between ETH and BTC it should be decoupled from the core API. Ideally the core API here just has a strategy based on the currency type of the wallet for which gateway it wants to forward the request to

packages/api/src/wallets/wallets.service.ts
Andre-Ibrahim: the post request doesn't contain that field so it's null at first

packages/api/test/unit/wallets/wallets.controller.spec.ts
Andre-Ibrahim: the return type of the create method is no longer user since `.save(user)` returns user but `.insert(user)` returns a InsertResult.

Andre-Ibrahim:
> I see you are trying to make 1 pr per subtask, its fine if you think it will be too large on its own but its a bit weird because half of this code will get deleted / moved when you do it in eth edge

do you think creating the wallet entity in the db should be done in eth-edge?
I thought it was the api's responsibility since it is not crypto specific.
Will refactor the code based on your sequence diagram

packages/api/src/wallets/wallets.service.ts
jason-gerard: It should contain what type of wallet it is though, the user has to select it in the UI

packages/api/test/unit/wallets/wallets.controller.spec.ts
jason-gerard: So instead of returning the result of insert just call the service method to find wallet (by address) or find user (by email) etc after insert has completed

PR #36, opened 2022-10-03
CRYP-79: create transaction db schema

alexbara2000:
Added a transaction table to our database


packages/common/src/entities/transaction.ts
```diff
@@ -0,0 +1,22 @@
...
+    wallet_in: string;
+
+    @Column({ nullable: false })
+    wallet_out: string;
+
+    @Column({ nullable: false })
+    amount: number;
+
+    @Column({ nullable: false })
+    password: string;
```
jason-gerard, line 18: What is this attribute for?

packages/common/src/entities/wallet.ts
```diff
@@ -15,4 +16,7 @@ export class Wallet {
     @ManyToOne(() => User, (user) => user.wallets)
     @JoinColumn({ name: "userId", referencedColumnName: "id" })
     user: User;
+
+    @ManyToMany(() => Transaction, (transaction) => transaction.wallet_in || transaction.wallet_out)
```
jason-gerard, line 20: Is it not one transaction to many wallets?

jason-gerard: It works with a many to many join table it's just not needed since the transaction is only associated with 2 wallet addresses not a variable number i.e. "many". So by one to many it really means constant to many or in this case two to many.

alexbara2000: In that case would it not make more sense to also update the ERD to be one to many?

jason-gerard: Yep makes sense could you update it in this pr? Just change this line https://github.com/SOEN-490-Capstone/Cryptify/blob/ed48623c364f4512d822f8271035f275dd32cbac/packages/diagrams/erd.puml#L60

I think the confusion with this, because i made an error in the diagram as well, comes from the fact that wallet_in and wallet_out are 2 distinct relationships not the same one, so I think it might actually make sense to have 2 relationships shown in the ERD
 `Wallet ||--o{ Transaction : receiver`
 `Wallet ||--o{ Transaction : sender`

And maybe add a note saying something along the lines of "there must be a wallet address for both in the transaction but only one of those wallets must be present in the wallets table in our system"

Let me know if this makes sense, should be more clear like this

alexbara2000: That makes a lot more sense and it keeps consistency between erd and the code.

packages/common/src/entities/wallet.ts
```diff
@@ -15,4 +16,10 @@ export class Wallet {
     @ManyToOne(() => User, (user) => user.wallets)
     @JoinColumn({ name: "userId", referencedColumnName: "id" })
     user: User;
+
+    @OneToMany(() => Transaction, (transaction) => transaction.wallet_in)
+    transactions_in: Transaction[];
+
+    @OneToMany(() => Transaction, (transaction) => transaction.wallet_out)
+    transactions_out: Transaction[];
```
jason-gerard, line 24: I would just leave it as you had it before with a sinlge attribute `transactions` but just update the relationship to be a one to many. The disjunction relationship you have `||` represents what the ERD shows with the 2 relationships.

```suggestion
    @OneToMany(() => Transaction, (transaction) => transaction.wallet_in || transaction.wallet_out)
    transactions: Transaction[];
```

jason-gerard (commented):
Also the attributes in the entities should be camelCase not snake_case as in the ERD

packages/common/src/entities/transaction.ts
```diff
@@ -0,0 +1,19 @@
...
+@Entity()
+export class Transaction {
+    @PrimaryColumn()
+    transaction_address: string;
+
+    @Column({ nullable: false })
+    wallet_in: string;
+
+    @Column({ nullable: false })
+    wallet_out: string;
```
jason-gerard, line 12: These should be changed to camelCasing like all the other attributes

jason-gerard (commented):
You still seem to have some commits outside of this pr included make sure to follow the steps I sent previously and force push

jason-gerard (approved):
Looks great nice work

PR #37, opened 2022-10-04
CRYP-27: sequence diagram for view wallets and update add wallets

jason-gerard:
## Changes
- Update the add wallets SD to include BtcEdge and only return after processing steps have been completed
- Add new SD for viewing wallets


domsec (approved):
Would we also need to show the `show progress screen` in the add wallet sequence?

If not then everything looks good and this PR can be merged.

jason-gerard:
> Would we also need to show the `show progress screen` in the add wallet sequence?
> 
> If not then everything looks good and this PR can be merged.

I think it makes sense to add, will update now.

domsec (approved):
One last thing...form error message(s) state

jason-gerard:
> One last thing...form error message(s) state

SD should only really show one flow, it gets too complicated if start doing the error handling, espcially the stuff thats just on the frontend. If we want to show case the actual states of the application an activity / state diagram would be much better, we can create some of those for complex flows that contain a lot of conditional states like add wallet while also keeping our SDs clean and to the point.

domsec:
> > One last thing...form error message(s) state
> 
> SD should only really show one flow, it gets too complicated if start doing the error handling, espcially the stuff thats just on the frontend. If we want to show case the actual states of the application an activity / state diagram would be much better, we can create some of those for complex flows that contain a lot of conditional states like add wallet while also keeping our SDs clean and to the point.

Sounds good.

PR #43, opened 2022-10-08
CRYP-62: Create backend API for getting a users wallets

Andre-Ibrahim:
- added GET users/:id/wallets endpoint to api returns a list of wallets
- refactored request gateway to support methods without body (get, delete)

### DIFF
--- a/packages/api/src/wallets/wallets.controller.ts
+++ b/packages/api/src/wallets/wallets.controller.ts
@@ -1,4 +1,4 @@
-import { Body, Controller, Post, UseGuards } from "@nestjs/common";
+import { Body, Controller, Get, Param, Post, UseGuards } from "@nestjs/common";
 import { CreateWalletRequest } from "@cryptify/common/src/requests/create_wallet_request";
 import { useValidate } from "@cryptify/common/src/hooks/use_validate";
 import { createWalletSchema } from "@cryptify/common/src/validations/create_wallet_schema";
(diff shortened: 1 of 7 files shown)


packages/common/src/validations/get_wallets_schema.ts
```diff
@@ -0,0 +1,5 @@
+import * as yup from "yup";
+
+export const getWalletsSchema = yup.object({
+    id: yup.number().required(),
```
jason-gerard, line 4: userId

packages/api/src/wallets/wallets.controller.ts
```diff
@@ -17,4 +20,12 @@ export class WalletsController {
         const createWalletReq = await useValidate(createWalletSchema, body);
         return this.walletsService.create(createWalletReq);
     }
+
+    @Get("/users/:id/wallets")
+    @UseGuards(JwtAuthGuard, AuthorizationGuard)
+    async findAll(@Param() param: GetWalletsRequest): Promise<Wallet[]> {
```
jason-gerard, line 26: params

jason-gerard (commented):
Good start couple things to refactor, let me know when the eth edge implémentions is done.

packages/common/src/validations/get_wallets_schema.ts
Andre-Ibrahim: This is the only way I could put the inferred type for params other than updating the path unless im missing something

Andre-Ibrahim:
> Good start couple things to refactor, let me know when the eth edge implémentions is done. 

Was going to do it in a different pr since theres two seperate jira tasks should i just do both here?

packages/common/src/validations/get_wallets_schema.ts
jason-gerard: ah i see never mind then. it's the only id we really have in the system anyways

jason-gerard: you can also rename the param when you pass it into the validation that would work. i generally don't like "id" because it's not very clear but up to you tbh

jason-gerard (approved):
I generally prefer to do things in 1 pr but if you like to do it in multiple no problem with me.

packages/common/src/validations/get_wallets_schema.ts
Andre-Ibrahim: Thought of that but then the type in the method signature would have to be another type or any

PR #44, opened 2022-10-09
CRYP-71: create eth edge api for getting users wallets

Andre-Ibrahim:
- added eth edge GET users/:id/wallets with alchemy integration
- refactored eth edge wallet service create method to check if the wallet exist in node before creating it in the database

### DIFF
--- a/packages/eth-edge/src/controllers/wallets.controller.ts
+++ b/packages/eth-edge/src/controllers/wallets.controller.ts
@@ -21,6 +21,6 @@ export class WalletsController {
     @Get("users/:id/wallets")
     async findAll(@Param() params: GetWalletsRequest): Promise<Wallet[]> {
         const getWalletReq = await useValidate(getWalletsSchema, params);
-        return await this.walletsService.findAll(getWalletReq.id);
+        return this.walletsService.findAll(getWalletReq.id);
     }
 }
--- a/packages/eth-edge/src/controllers/wallets.controller.ts
+++ b/packages/eth-edge/src/controllers/wallets.controller.ts
@@ -1,9 +1,12 @@
-import { Body, Controller, Post } from "@nestjs/common";
+import { Body, Controller, Param, Post } from "@nestjs/common";
 import { CreateWalletRequest } from "@cryptify/common/src/requests/create_wallet_request";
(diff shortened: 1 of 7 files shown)


packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -21,14 +25,23 @@ export class WalletsService {
...
-        await this.walletRepository.insert(reqWallet);
+        let balance;
+        try {
+            balance = await this.alchemyNodeService.getBalance(createWalletReq.address);
+        } catch (error) {
+            throw new BadRequestException(ERROR_WALLET_DOES_NOT_EXIST(CurrencyType.ETHEREUM));
+        }
 
-        const balance = await this.alchemyNodeService.getBalance(createWalletReq.address);
+        await this.walletRepository.insert(reqWallet);
```
jason-gerard, line 35: Can you leave this as it was, I refactored it in my PR, we can't actually know if a wallet doesn't exist or not because a wallet can exist but just never have been transacted with, so in this case I just return "0". That error message can also be removed since it will never occur

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -14,4 +17,10 @@ export class WalletsController {
         const createWalletReq = await useValidate(createWalletSchema, body);
         return this.walletsService.create(createWalletReq);
     }
+
+    @Get("users/:id/wallets")
+    async findAll(@Param() params: GetWalletsRequest): Promise<Wallet[]> {
+        const getWalletReq = await useValidate(getWalletsSchema, params);
+        return await this.walletsService.findAll(getWalletReq.id);
```
jason-gerard, line 24: Don't need to await when you return a value, it messes with nodes event loop and the ordering of things, best practice is to let the caller await the value since the function is wrapping it in a promise again anyways

```suggestion
        return this.walletsService.findAll(getWalletReq.id);
```

jason-gerard (approved):

packages/eth-edge/src/services/wallets.service.ts
Andre-Ibrahim: ah thought 0 transaction would just have a balance of 0. Makes sense i removed it!

PR #45, opened 2022-10-10
CRYP-69: Ethereum wallet transactions backfilling

alexbara2000:
Added transaction entity and backfilled them once a new account is added

### DIFF
--- a/packages/common/src/domain/entities/transaction.ts
+++ b/packages/common/src/domain/entities/transaction.ts
@@ -11,8 +11,8 @@ export class Transaction {
     @Column({ nullable: false })
     walletOut: string;
 
-    @Column({ nullable: false, type: "double precision" })
-    amount: number;
+    @Column({ nullable: false})
+    amount: string;
 
     @CreateDateColumn({ type: "timestamp" })
     createdAt: Date;
--- a/packages/common/src/domain/entities/transaction.ts
+++ b/packages/common/src/domain/entities/transaction.ts
@@ -11,7 +11,7 @@ export class Transaction {
     @Column({ nullable: false })
     walletOut: string;
 
-    @Column({ nullable: false})
+    @Column({ nullable: false })
     amount: string;
 
     @CreateDateColumn({ type: "timestamp" })
--- a/packages/common/src/domain/entities/transaction.ts
+++ b/packages/common/src/domain/entities/transaction.ts
@@ -11,7 +11,7 @@ export class Transaction {
     @Column({ nullable: false })
     walletOut: string;
 
-    @Column({ nullable: false })
+    @Column({ nullable: false, type: "double precision" })
     amount: number;
 
     @CreateDateColumn({ type: "timestamp" })
--- a/packages/common/src/errors/error_messages.ts
+++ b/packages/common/src/errors/error_messages.ts
@@ -8,3 +8,6 @@ export const ERROR_WALLET_ALREADY_ADDED_TO_ACCOUNT = (currency: string) =>
 export const ERROR_WALLET_NAME_ALREADY_ADDED_TO_ACCOUNT = "name:This name is already associated with another wallet.";
 
 export const ERROR_NOP = "nop";
+export const ERROR_WALLET_ADDRESS_FOR_CURRENCY = (currency: string) =>
+    `This wallet address is already associated with another ${currency} wallet.`;
+export const ERROR_TRANSACTION_ALREADY_ADDED = "A duplicate transaction has been found and the bacfil cannot be completed.";
--- a/packages/common/src/errors/error_messages.ts
+++ b/packages/common/src/errors/error_messages.ts
@@ -8,6 +8,3 @@ export const ERROR_WALLET_ALREADY_ADDED_TO_ACCOUNT = (currency: string) =>
 export const ERROR_WALLET_NAME_ALREADY_ADDED_TO_ACCOUNT = "name:This name is already associated with another wallet.";
 
 export const ERROR_NOP = "nop";
-export const ERROR_WALLET_ADDRESS_FOR_CURRENCY = (currency: string) =>
-    `This wallet address is already associated with another ${currency} wallet.`;
-export const ERROR_TRANSACTION_ALREADY_ADDED = "A duplicate transaction has been found and the bacfil cannot be completed.";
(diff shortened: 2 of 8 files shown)


packages/common/src/errors/error_messages.ts
```diff
@@ -4,3 +4,4 @@ export const ERROR_NOP = "nop";
 export const ERROR_WALLET_ADDRESS_FOR_CURRENCY = (currency: string) =>
     `This wallet address is already associated with another ${currency} wallet.`;
 export const ERROR_WALLET_ALREADY_ADDED_TO_ACCOUNT = "This wallet address is already associated with another wallet.";
+export const ERROR_TRANSACTION_ALREADY_ADDED = "A duplicate transaction has been found and the bacfil cannot be completed.";
```
domsec, line 7: Is this an error message that will be displayed to the user?

domsec (commented):
I just want to confirm that the backfilling does indeed occur during the `Adding a Wallet` loading screen.

packages/common/src/domain/entities/transaction.ts
```diff
@@ -11,7 +11,7 @@ export class Transaction {
     @Column({ nullable: false })
     walletOut: string;
 
-    @Column({ nullable: false })
+    @Column({ nullable: false, type: "double precision" })
     amount: number;
```
jason-gerard, line 15: This should be a string we can't trust javascript floating points

packages/eth-edge/src/modules/wallets.module.ts
```diff
@@ -4,10 +4,14 @@ import { TypeOrmModule } from "@nestjs/typeorm";
...
 import { AlchemyNodeService } from "@cryptify/eth-edge/src/services/alchemy_node.service";
+import { TransactionsService } from "@cryptify/eth-edge/src/services/transactions.service";
+import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
+import { TransactionModule } from "./transactions.module";
 
 @Module({
-    imports: [TypeOrmModule.forFeature([Wallet])],
+    imports: [TypeOrmModule.forFeature([Wallet]), TransactionModule],
     controllers: [WalletsController],
+    exports: [WalletsService],
```
jason-gerard, line 14: I dont think this needs to be exported

```suggestion
```

packages/eth-edge/src/services/alchemy_node.service.ts
```diff
@@ -16,4 +16,30 @@ export class AlchemyNodeService {
...
+            fromAddress: wallet,
+            excludeZeroValue: true,
+            category: [AssetTransfersCategory.EXTERNAL],
+            withMetadata: true
+            });
+    }
+
+    async getBlock(blockNum: string) {
+        return await this.alchemy.core.getBlock(blockNum);
+    }
```
jason-gerard, line 44: We can delete this since it itsn't needed to get the timestamp with the metadata flag now
```suggestion
```

packages/common/src/errors/error_messages.ts
alexbara2000: I created this error message from a backend perspective, I did not think about frontend implications. Did we plan on showing an error message to the user if the backfilling fails?

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -25,6 +27,7 @@ export class WalletsService {
 
         const balance = await this.alchemyNodeService.getBalance(createWalletReq.address);
         const wallet = await this.findOne(createWalletReq.address, createWalletReq.userId);
+        this.transactionsService.backfil(wallet.address);
```
jason-gerard, line 30: Backfill is spelled with 2 l's, method needs to be awaited or else it is just scheduled in the event queue, and it should be moved up. You can consult the sequence diagram but it should be after gettng the balance but before getting the wallet.
```suggestion
        await this.transactionsService.backfill(wallet.address);
```

packages/common/src/errors/error_messages.ts
alexbara2000: Also, yes the backfilling does occur during the adding of a wallet loading screen

packages/eth-edge/src/services/transactions.service.ts
```diff
@@ -0,0 +1,51 @@
...
+            }
+            transArr.push(transaction) 
+        }
+        try{
+            const reqtransaction = this.transactionRepository.create(transArr);
+            await this.transactionRepository.insert(reqtransaction);
+        }
+        catch{
+            throw new BadRequestException(ERROR_TRANSACTION_ALREADY_ADDED);
+        }
```
jason-gerard, line 48: You can use the `save` method instead of the `insert` method which will act as an upsert. This is fine for us since a duplicate wallet will just be skipped over / updated with the same values.

We really don't want to throw an error here under any circumstances since it will completely break the add to wallet process. Throwing here will cancel the rest of the steps without rolling back any other state changes made.

```suggestion
        const transactions = this.transactionRepository.create(transArr);
        await this.transactionRepository.save(transactions);
```

packages/common/src/errors/error_messages.ts
jason-gerard: Backfilling doesn't actually occur during the wallet loading screen right now because the method is async and not awaited.

alexbara2000: I thought the loading screen display was handled by the frontend until the result of the call is returned.

packages/common/src/domain/entities/transaction.ts
alexbara2000: The alchemy object returns a "number" but I can change it to a string if you want

packages/common/src/errors/error_messages.ts
jason-gerard: It is but you are just basically creating a new thread and then proceeding / returning without waiting for the thread to complete, it isn't really a thread but the idea is similar

domsec: > I created this error message from a backend perspective, I did not think about frontend implications. Did we plan on showing an error message to the user if the backfilling fails?

Not that I'm aware of. I figured we'd silently fail/ignore them as @jason-gerard mentioned.

packages/common/src/domain/entities/transaction.ts
jason-gerard: It doesn't actually return a number if you check the implementation, JS number can only store 17 decimals it would work. It actually uses the BigNumber data type to properly keep the value. We could also use it but then we would have issues on the frontend and with serializing / deserializing it over the wire.

jason-gerard (approved):

jason-gerard (approved):

jason-gerard (approved):

PR #46, opened 2022-10-11
CRYP-70: implement alchemy webhook to get new transactions

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-70

## Changes
- Add webhook API route for alchemy
- Implement event handler for webhook
- Parallelize calls in add wallet
- Add alchemy gateway with update webhook address call


packages/eth-edge/src/services/transactions.service.ts
```diff
@@ -41,7 +43,30 @@ export class TransactionsService {
...
+        // then convert the remaining activities into transaction entities
+        const transactions = addressActivityEvent.event.activity
+            .filter((activity) => activity.category === AssetTransfersCategory.EXTERNAL && activity.asset === "ETH")
+            .map((activity) =>
+                this.transactionsRepository.create({
+                    transactionAddress: activity.hash,
+                    walletIn: activity.toAddress,
+                    walletOut: activity.fromAddress,
+                    amount: activity.value.toString(),
+                    createdAt: new Date(addressActivityEvent.createdAt),
```
alexbara2000, line 61: I am surprised you had to transform the createdAt field into a Date since I did not have to do it for the backfill. Are you sure it is not already a Date object?

alexbara2000 (approved):
Looks good

packages/eth-edge/src/services/transactions.service.ts
jason-gerard: You are using the SDK so it handles all the over the wire type conversions (cause its only strings, numbers, and byte data) so it did the string to date conversion for you, my request is coming straight from an HTTP request where nothing is type casted. So that `createdAt` field is just a ISO string.

PR #47, opened 2022-10-11
CRYP-63: implement frontend for viewing wallets

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-27

## Changes
- Implement view wallets screen
- Implement navigation to view wallets from settings and add wallet
- Update getBalance to return ETHER instead of WEI
- Update find all wallet api to return wallets and balances

## Screenshots
image
image
image
image


packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+        borderColor: "#E5E5E5",
+    },
+    walletName: {
+        fontSize: 17,
+        lineHeight: 23,
+    },
+    walletAddress: {
+        fontSize: 13,
+        lineHeight: 17,
+        color: "#A3A3A3",
```
domsec, line 155: Add color to the component with the associated name instead of here.

packages/client/screens/ViewWalletsScreen.tsx
```diff
@@ -1,6 +1,62 @@
...
+    );
 }
+
+const styles = StyleSheet.create({
+    view: {
+        flex: 1,
+    },
+    scrollView: {
+        paddingHorizontal: 15,
+        paddingTop: 10,
```
domsec, line 52: Change to 30

packages/client/screens/ViewWalletsScreen.tsx
```diff
@@ -1,6 +1,62 @@
...
+            )}
+        </View>
+    );
 }
+
+const styles = StyleSheet.create({
+    view: {
+        flex: 1,
+    },
+    scrollView: {
```
domsec, line 50: Add padding bottom of 15

packages/client/components/icons/faChevronDownCustom.tsx
```diff
@@ -0,0 +1,18 @@
...
+        // SVG viewbox width (in pixels)
+        512,
+        // SVG viewbox height (in pixels)
+        512,
+        // Aliases (not needed)
+        [],
+        // Unicode as hex value (not needed)
+        "",
+        // SVG path data
+        "M233.4 406.6c12.5 12.5 32.8 12.5 45.3 0l192-192c12.5-12.5 12.5-32.8 0-45.3s-32.8-12.5-45.3 0L256 338.7 86.6 169.4c-12.5-12.5-32.8-12.5-45.3 0s-12.5 32.8 0 45.3l192 192z",
```
domsec, line 14: Can you just verify this is indeed the path for chevron down in regular font weight. I couldn't find this path in any of the fontawesome files we're extracting this info from.

domsec (commented):
This is my preliminary code review since I cannot as of yet fully test everything because my local dev environment is still broken. There is most definitely a few things I might have missed within this review. A second pass will most likely be needed once these initial findings are resolved.

packages/eth-edge/src/services/alchemy_node.service.ts
```diff
@@ -18,13 +19,14 @@ export class AlchemyNodeService {
...
-            return balance.toString();
+            // any calculations with the balance, we will also convert the WEI value
+            // to ETHER since that is standard for EthEdge
+            return Web3.utils.fromWei(balance.toString(), "ether");
         } catch (error) {
             // If there is any error, mainly if the wallet is not found return a balance of 0
             // this is done because we can't verify that the wallet doesn't exist just because
             // it doesn't have any previous transactions
-            return "0";
+            return "0.000000000000000000";
```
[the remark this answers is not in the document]
jason-gerard: These are stored values and will always have 18 decimals, how we display it to the user can change independently of that it just gives us the most sig figs if that makes sense.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+                                activeSections={activeSections[currency.type]}
+                                renderHeader={renderHeader}
+                                renderContent={renderContent}
+                                onChange={(activeSections) =>
+                                    setActiveSections((curr) => ({
+                                        ...curr,
+                                        [currency.type]: activeSections,
+                                    }))
+                                }
+                                underlayColor="#E5E5E5"
```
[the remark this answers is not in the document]
jason-gerard: It doesn't work because this isnt a native base component, they removed the typings for it in v3 so even if you import it you cant use it.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+    },
+    chevronIcon: {
+        color: "#A3A3A3",
+        marginLeft: "auto",
+        marginRight: 5,
+    },
+    walletItemWrapper: {
+        borderWidth: 1,
+        borderColor: "#E5E5E5",
+        borderTopWidth: 0,
```
domsec, line 139: Could change this 1 and have the items handle the divider instead of calculating the last one. See above comment.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+    },
+    ethereumIcon: {
+        color: "#3C3C3D",
+    },
+    chevronIcon: {
+        color: "#A3A3A3",
+        marginLeft: "auto",
+        marginRight: 5,
+    },
+    walletItemWrapper: {
```
domsec, line 136: Missing vertical padding of 10?

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    <Box
+                        key={i}
+                        style={{
+                            ...styles.walletItemWrapper,
+                            borderBottomLeftRadius: i === walletsByType[currency.type].length - 1 ? 10 : 0,
+                            borderBottomRightRadius: i === walletsByType[currency.type].length - 1 ? 10 : 0,
+                            borderBottomWidth: i === walletsByType[currency.type].length - 1 ? 1 : 0,
+                        }}
+                    >
+                        <HStack style={styles.walletItem} alignItems="center">
```
domsec, line 68: There should be a 5px space between the wallet name/address and balance. That way we don't have a situation where the balance and wallet name are extremely close.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+        borderWidth: 1,
+        borderColor: "#E5E5E5",
+        borderTopWidth: 0,
+        borderBottomWidth: 0,
+        paddingHorizontal: 15,
+    },
+    walletItem: {
+        paddingVertical: 15,
+        borderTopWidth: 1,
+        borderColor: "#E5E5E5",
```
domsec, line 146: Add border color style to the component with the associated color name instead of here.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        style={{
+                            ...styles.walletItemWrapper,
+                            borderBottomLeftRadius: i === walletsByType[currency.type].length - 1 ? 10 : 0,
+                            borderBottomRightRadius: i === walletsByType[currency.type].length - 1 ? 10 : 0,
+                            borderBottomWidth: i === walletsByType[currency.type].length - 1 ? 1 : 0,
+                        }}
+                    >
+                        <HStack style={styles.walletItem} alignItems="center">
+                            <VStack>
+                                <Text style={styles.walletName}>{wallet.name}</Text>
```
domsec, line 70: Make sure the wallet name wraps to a new line if too long.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+                                <Box marginTop="2px"></Box>
+                                <Text style={styles.walletAddress}>{formatWalletAddress(wallet.address)}</Text>
+                            </VStack>
+                            <Text style={styles.walletBalance}>
+                                {wallet.balance} {currency.currencyTag}
```
domsec, line 75: Just a note we'll have to refactor the balance to take into account the new specs regarding large balances (i.e. ethereum with 18 trailing digits).

domsec (commented):
Second pass done. All comments can now begin to be resolved. 

One thing to note, from what I can see the accordion items don't seem to be responsive from a code perspective (I have yet to fully test this out). Just make sure the appropriate css from Figma is there. All UI mockups are built with responsiveness in mind so the associated Figma code has such code present under the `Inspect` right sidebar.

packages/client/components/WalletsListAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+    const renderHeader = (currency: CurrencyDisplayData, _: number, isActive: boolean) => {
+        return (
+            <HStack
+                height="66px"
+                alignItems="center"
+                style={{
+                    ...styles.header,
+                    borderBottomRightRadius: isActive ? 0 : 10,
+                    borderBottomLeftRadius: isActive ? 0 : 10,
```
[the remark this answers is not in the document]
Andre-Ibrahim: small logic like this is fine to be duplicate since its easy to understand and the dev doesn't have to go outside the jsx to see what the variable is actually doing, same applies for the other comments in my opinion.

domsec (approved):

packages/api/src/wallets/wallets.service.ts
```diff
@@ -26,7 +26,7 @@ export class WalletsService {
         return this.ethEdgeGateway.createWallet(req);
     }
 
-    async findAll(req: GetWalletsRequest): Promise<Wallet[]> {
+    async findAll(req: GetWalletsRequest): Promise<WalletWithBalance[]> {
```
Andre-Ibrahim, line 29: this doesn't follow the convention but since there is no reason for us to query the wallets without the balance it makes sense to me

packages/common/src/helpers/function_utils.ts
```diff
@@ -0,0 +1,3 @@
+export function zip<T, U>(a: T[], b: U[]): [T, U][] {
```
Andre-Ibrahim, line 1: No clue what is going on here, can you please explain

Andre-Ibrahim (approved):
just one question for my understanding, other than that looks good to me

packages/common/src/helpers/function_utils.ts
jason-gerard: This is a standard implementation of a typed zip function which takes two list and combines the nth element of each list into a tuple in the new list. Very standard FP technique.

packages/api/src/wallets/wallets.service.ts
jason-gerard: WalletWithBalance is a Wallet :0 therefore convention === followed

Andre-Ibrahim: more like convention == followed

jason-gerard: I see what you did there, I'm actually really impressed

PR #48, opened 2022-10-12
CRYP-23: Implement signout button and functionality on settings page

PolaFarid:
unknown

### DIFF
--- a/packages/client/components/SignOutButton.tsx
+++ b/packages/client/components/SignOutButton.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+import { Button } from "native-base";
+import { StyleSheet } from "react-native";
+import StorageService from "../services/storage_service";
+import { KEY_JWT } from "../constants/storage_keys";
+
+export default function SignOutButton() {
+    async function handleSignOut(): Promise<void> {
+        // TODO refactor this function to redirect user to welcome screen
+        // this will be done automatically when the token is removed
+        // after refactoring the navigation which will be done in Sprint 3
+        // when the entire sign up / sign in flow is completed
+        await StorageService.remove(KEY_JWT);
+    }
+
+    return (
+        <Button variant="outline" _text={styles.signOutButtonText} onPress={handleSignOut}>
+            Sign out
+        </Button>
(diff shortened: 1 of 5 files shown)


jason-gerard (approved):
Looks good, nice work.

packages/client/theme/components/button.ts
```diff
@@ -24,6 +24,19 @@ export const button = {
                 opacity: "60",
             },
         },
+        outline: {
+            bg: "white",
+            borderColor: "text.200",
+            _text: {
+                color: "error.500",
```
domsec, line 31: Set the default text color "text.700" for outline variant buttons. Specific text colors should be set per Button component instance.

packages/client/components/SignOutButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+    async function handleSignOut(): Promise<void> {
+        // TODO refactor this function to redirect user to welcome screen
+        // this will be done automatically when the token is removed
+        // after refactoring the navigation which will be done in Sprint 3
+        // when the entire sign up / sign in flow is completed
+        await StorageService.remove(KEY_JWT);
+    }
+
+    return (
+        <Button variant="outline" _text={styles.signOutButtonText} onPress={handleSignOut}>
```
domsec, line 17: Replace `_text={styles.signOutButtonText}` with `style={styles.signOutButton}` and add

```
_text={{
      color: "error.500"
    }}
```

domsec (commented):
Re-submit a PR with these changes.

packages/client/components/SignOutButton.tsx
Andre-Ibrahim: Button style prop doesnt take a StyleSheet object like that

Andre-Ibrahim: `Type '{ fontSize: number; lineHeight: number; }' is not assignable to type 'StyleProp<ViewStyle> | ((state: PressableStateCallbackType) => StyleProp<ViewStyle>)'.ts(2322)
`

PR #50, opened 2022-10-12
CRYP-114: refactored folder structure for api to be by type instead of by domain

Andre-Ibrahim:


jason-gerard:
Looks good but can you double check in your ide to make sure all the paths are good, should give a red underline error if any are bad.

jason-gerard (approved):

Andre-Ibrahim:
> Looks good but can you double check in your ide to make sure all the paths are good, should give a red underline error if any are bad.

all good on my end, also wouldn't the CI fail?

jason-gerard:
> > Looks good but can you double check in your ide to make sure all the paths are good, should give a red underline error if any are bad.
> 
> 
> 
> all good on my end, also wouldn't the CI fail?

Only if the path is imported as part of the CI like if the file is used. JS is weird in that it doesn't know if paths are good or not until it runs that line. A linter wont catch it for example.

PR #53, opened 2022-10-19
CRYP-20: Setup detox system testing framework

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-20


packages/client/app.json
```diff
@@ -1,33 +1,41 @@
...
+      "backgroundColor": "#ffffff"
+    },
+    "updates": {
+      "fallbackToCacheTimeout": 0
+    },
+    "assetBundlePatterns": [
+      "**/*"
+    ],
+    "ios": {
+      "supportsTablet": true
```
domsec, line 22: Should we set this to false since we aren't currently supporting tablet?

domsec (approved):
Looks good.

packages/client/app.json
jason-gerard: Yep there is a lot of other config we will do in this file for iOS and android once we start loading it on actual devices and seeing how it works, couple other things I'll do in this pr before merging though, update docs and set up some scripts mostly

domsec: Sounds good.

Andre-Ibrahim (approved):
very nice

PR #55, opened 2022-10-23
CRYP-116: implement base abstract gateway

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-116


Andre-Ibrahim (approved):
I like it

PR #59, opened 2022-10-27
CRYP-74: implement backend api for getting a users transactions

alexbara2000:
Added ability to get transactions for a user in the eth architecture

### DIFF
--- a/packages/api/src/services/transactions.service.ts
+++ b/packages/api/src/services/transactions.service.ts
@@ -1,0 +1,21 @@
+import { Injectable } from "@nestjs/common";
+import { GetTransactionsRequest } from "@cryptify/common/src/requests/get_transaction_request";
+import { EdgeGatewayStrategyFactory } from "@cryptify/api/src/gateways/edge-gateway/edge_gateway_strategy_factory";
+import { CurrencyType } from "@cryptify/common/src/domain/currency_type";
+import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
+
+@Injectable()
+export class TransactionsService {
+    constructor(private readonly edgeGatewayStrategyFactory: EdgeGatewayStrategyFactory) {}
+
+    async findAll(req: GetTransactionsRequest): Promise<Transaction[]> {
+        // Map through the currency types to build a list of strategies, then map through each strategy
+        // and get the transactions from the associated edge service in parallel, finally flatten the 2D
+        // array of wallets into a single array
+        const strategies = Object.keys(CurrencyType).map((type) =>
+            this.edgeGatewayStrategyFactory.get(CurrencyType[type]),
+        );
+        const transactionsByType = await Promise.all(strategies.map((strategy) => strategy.getTransactions(req)));
+        return transactionsByType.flat();
+    }
+}
--- a/packages/api/src/services/transactions.service.ts
+++ b/packages/api/src/services/transactions.service.ts
@@ -9,9 +9,6 @@ export class TransactionsService {
     constructor(private readonly edgeGatewayStrategyFactory: EdgeGatewayStrategyFactory) {}
 
     async findAll(req: GetTransactionsRequest): Promise<Transaction[]> {
-        // Map through the currency types to build a list of strategies, then map through each strategy
-        // and get the transactions from the associated edge service in parallel, finally flatten the 2D
-        // array of wallets into a single array
         const strategies = Object.keys(CurrencyType).map((type) =>
             this.edgeGatewayStrategyFactory.get(CurrencyType[type]),
         );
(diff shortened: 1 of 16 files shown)


packages/api/src/services/transactions.service.ts
```diff
@@ -0,0 +1,21 @@
...
+import { EdgeGatewayStrategyFactory } from "@cryptify/api/src/gateways/edge-gateway/edge_gateway_strategy_factory";
+import { CurrencyType } from "@cryptify/common/src/domain/currency_type";
+import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
+
+@Injectable()
+export class TransactionsService {
+    constructor(private readonly edgeGatewayStrategyFactory: EdgeGatewayStrategyFactory) {}
+
+    async findAll(req: GetTransactionsRequest): Promise<Transaction[]> {
+        // Map through the currency types to build a list of strategies, then map through each strategy
```
jason-gerard, line 12: Delete comment

packages/common/src/requests/get_transaction_request.ts
```diff
@@ -0,0 +1,4 @@
+import { InferType } from "yup";
+import { getTransactionsSchema } from "../validations/get_transactions_schema";
+
+export type GetTransactionsRequest = InferType<typeof getTransactionsSchema>;
```
jason-gerard, line 4: should be an interface

jason-gerard: Nevermind, I guessed none of the other ones were changed

packages/eth-edge/src/services/transactions.service.ts
```diff
@@ -52,4 +55,12 @@ export class TransactionsService {
         // has already been processed by the system
         await this.transactionsRepository.save(transactions);
     }
+
+    async findAll(userId: number): Promise<Transaction[]> {
+        const userWallets = await this.walletsService.findAll(userId);
+        const addresses = userWallets.map((t) => t.address.toLowerCase());
```
jason-gerard, line 61: We should make sure to also lower case all addresses in general when we insert?
```suggestion
        const addresses = userWallets.map((transaction) => transaction.address.toLowerCase());
```

packages/api/src/services/transactions.service.ts
```diff
@@ -0,0 +1,21 @@
...
+        // and get the transactions from the associated edge service in parallel, finally flatten the 2D
+        // array of wallets into a single array
```
jason-gerard, line 14: Small typo in comment

```suggestion
        // Map through the currency types to build a list of strategies, then map through each strategy
        // and get the transactions from the associated edge service in parallel, finally flatten the 2D
        // array of transactions into a single array
```

packages/eth-edge/src/controllers/transactions.controller.ts
```diff
@@ -13,4 +17,10 @@ export class TransactionsController {
         this.transactionsService.handleAddressActivityEvent(body);
         return;
     }
+
+    @Get("users/:id/transactions")
+    async findAll(@Param() params: GetTransactionsRequest): Promise<Transaction[]> {
+        const getTransactionReq = await useValidate(getTransactionsSchema, params);
+        return this.transactionsService.findAll(getTransactionReq.id);
```
jason-gerard, line 24: Ideally the services would take the request object like what is done in the api but i noticed the other `findAll` are done like this so its fine for now

```suggestion
        return this.transactionsService.findAll(getTransactionReq);
```

packages/eth-edge/src/services/transactions.service.ts
```diff
@@ -52,4 +55,12 @@ export class TransactionsService {
         // has already been processed by the system
         await this.transactionsRepository.save(transactions);
     }
+
+    async findAll(userId: number): Promise<Transaction[]> {
+        const userWallets = await this.walletsService.findAll(userId);
```
jason-gerard, line 60:
```suggestion
        const wallets = await this.walletsService.findAll(userId);
```

jason-gerard (approved):

packages/common/src/requests/get_transaction_request.ts
alexbara2000: get_wallet_request is not an interface, should I change both to be interfaces?

packages/client/gateways/transactions_gateway.ts
```diff
@@ -0,0 +1,20 @@
+import { JwtToken } from "@cryptify/common/src/domain/jwt_token";
+import { GetTransactionsRequest } from "@cryptify/common/src/requests/get_transaction_request";
+import { AbstractApiGateway } from "./abstract_api_gateway";
+import { Method } from "@cryptify/common/src/utils/gateway/abstract_gateway";
+import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
+
+export class TransactionsGateway extends AbstractApiGateway {
```
jason-gerard, line 7: Not sure this belongs in this pull request

PR #60, opened 2022-10-27
CRYP-100: welcome screen

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-100

## Changes
- Add welcome screen when user is signed out
- Implement signed in and signed out flow
- Add back buttons on sign in and sign up screens
- Expose jwt token as context state to all screens


packages/client/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export default function WelcomeScreen({ navigation }: GuestStackScreenProps<"WelcomeScreen">) {
+    return (
+        <View style={styles.view}>
+            <Box safeArea></Box>
+            <Box style={styles.titleContainer}>
+                <Logo />
+                <Box marginTop="20px"></Box>
+                <Text style={styles.subtext}>Crypto asset management{"\n"}made simple</Text>
+            </Box>
```
domsec, line 17: Wouldn't it be better to refactor this with `VStack` similar to what we did in Sign In/Sign Up screens?

packages/client/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+            <Button onPress={() => navigation.navigate("SignUpScreen")}>Sign up</Button>
+            <Box marginTop="15px"></Box>
+            <Button variant="outline" onPress={() => navigation.navigate("SignInScreen")}>
+                Sign in
+            </Button>
+            <Box marginBottom="20px"></Box>
```
domsec, line 23: Same as above comment.

domsec (approved):
Besides my two refactor comments, everything looks good. I'm really glad the welcome screen background alignment is a 1 to 1 match with the mockups.

packages/client/screens/WelcomeScreen.tsx
jason-gerard: yep you are right, updated

packages/client/screens/WelcomeScreen.tsx
jason-gerard: updated

domsec (approved):
Great work.

PR #63, opened 2022-10-28
CRYP-101: home page

jason-gerard:

## Changes
- Support the user object in the app context
- Add home page header and content
- Extract `WalletsList` component
- Add new `CurrencyAmount` component
- Add `currency_service`


packages/client/components/CurrencyAmount.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type Props = {
+    currency: CurrencyDisplayData;
+    amount: string;
+    amountStyles: StyleProp<TextStyle>;
+    currencyCodeStyles: StyleProp<TextStyle>;
+};
+
+export function CurrencyAmount({ currency, amount, amountStyles, currencyCodeStyles }: Props) {
+    const formattedAmount = currency.type == CurrencyType.ETHEREUM ? Web3.utils.fromWei(amount, "ether") : amount;
```
domsec, line 16: Add commas to separate thousands.

packages/client/components/wallets-list/WalletsListAccordion.tsx
```diff
@@ -65,15 +79,18 @@ export function WalletsListAccordion({ wallets }: Props) {
                             borderBottomWidth: i === walletsByType[currency.type].length - 1 ? 1 : 0,
                         }}
                     >
-                        <HStack style={styles.walletItem} alignItems="center">
+                        <HStack style={styles.walletItem} alignItems="center" space="5px">
                             <VStack>
```
domsec, line 83: We could add a space property of 2 and not need the `<Box marginTop="2px"></Box>` underneath

packages/client/screens/ViewWalletsScreen.tsx
```diff
@@ -1,43 +1,14 @@
...
-                <ScrollView style={styles.scrollView}>
-                    <WalletsListAccordion wallets={wallets} />
-                </ScrollView>
-            ) : (
-                <Center alignItems="center" marginY="auto">
-                    <FontAwesomeIcon icon={faWalletCustom} style={styles.walletIcon} size={56} />
-                    <Text style={styles.noWalletsText}>You do not have any wallets.</Text>
-                </Center>
-            )}
+            <Box marginTop="10px"></Box>
```
domsec, line 10: I don't think we need this over here since it would cause the `You do not have any wallets` contained to no longer be centered. It's safe from what I can see to increase the `walletsListTitleContainer` marginTop property to 30.

domsec (approved):
Few minor things. The rest looks good.

packages/client/components/CurrencyAmount.tsx
jason-gerard: done

packages/client/screens/ViewWalletsScreen.tsx
jason-gerard: This screen is used from the settings page, so removing the margin top here would make it too close to the top. What I went ahead and did was add a negative margin in the 'no wallets' component which will bump it back up 10px when that view is active making it centered in the settings page and home page flow

domsec: I see. We could refactor the settings page to not require this hack but for now this works with me.

domsec (approved):
Looks good

PR #64, opened 2022-10-28
Cryp-28: list of all transactions associated to each of my eth wallets

alexbara2000:
Added transaction list with the ability to display date separators

### DIFF
--- a/packages/client/components/Transaction.tsx
+++ b/packages/client/components/Transaction.tsx
@@ -23,7 +23,7 @@ export function Transaction({ transaction, walletAddress }: Props) {
     }
 
     function formatAmount(amount: string): string {
-        return amount.length>15 ? amount.substring(0,15)+"...": amount;
+        return amount.length > 15 ? amount.substring(0, 15) + "..." : amount;
     }
 
     function getFromattedDate(date: Date): string {
--- a/packages/client/components/Transaction.tsx
+++ b/packages/client/components/Transaction.tsx
@@ -22,6 +22,10 @@ export function Transaction({ transaction, walletAddress }: Props) {
         return transaction.transactionAddress.substring(0, 2) == "0x" ? "ETH" : "BTC";
     }
 
+    function formatAmount(amount: string): string {
+        return amount.length>15 ? amount.substring(0,15)+"...": amount;
+    }
+
     function getFromattedDate(date: Date): string {
         const day = date.getDay();
         const year = date.getFullYear();
@@ -51,7 +55,7 @@ export function Transaction({ transaction, walletAddress }: Props) {
                             style={isIncommingTransaction ? styles.transactionAmountIn : styles.transactionAmountOut}
                         >
                             {isIncommingTransaction ? "+" : "-"}
-                            {transaction.amount}
+                            {formatAmount(transaction.amount)}
                         </Text>
                     </HStack>
                     <HStack>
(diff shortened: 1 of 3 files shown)


alexbara2000:
image
image

packages/client/components/Transaction.tsx
```diff
@@ -22,6 +22,10 @@ export function Transaction({ transaction, walletAddress }: Props) {
         return transaction.transactionAddress.substring(0, 2) == "0x" ? "ETH" : "BTC";
     }
 
+    function formatAmount(amount: string): string {
+        return amount.length > 15 ? amount.substring(0, 15) + "..." : amount;
+    }
```
jason-gerard, line 27: There is a func `getFormattedAmount` in `currency_service.ts` that does this properly and then for the elipses just use the prop `isTruncated` on the `Text` tag this way its dynamic.

packages/client/components/TransactionList.tsx
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import { Text, Box } from "native-base";
+import { StyleSheet } from "react-native";
+import { Transaction as TransactionEntity } from "@cryptify/common/src/domain/entities/transaction";
+import { Transaction } from "./Transaction";
```
jason-gerard, line 5: Rename this file to a better name that doesnt conflict with the entity

```suggestion
import { TransactionListItem } from "./TransactionListItem";
```

jason-gerard (approved):
Logic is a bit clunky, but pretty impressed good work.

packages/client/components/TransactionList.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+type Props = {
+    transactions: Transaction[];
+    walletAddress: string;
+    displaySeparation: boolean;
+};
+
+export function TransactionList({ transactions, walletAddress, displaySeparation }: Props) {
+    let savedDate = new Date();
+    function renderSeparation(date: Date) {
```
domsec, line 15: Just a personal preference, but something on the lines of `renderHeader` or something along those lines would be more descriptive.

packages/client/components/TransactionList.tsx
```diff
@@ -0,0 +1,69 @@
...
+                <Text color="text.500" style={styles.dateSeparator}>
+                    {date.toLocaleString("en-US", { month: "long" }) + " " + date.getFullYear()}
+                </Text>
+            </Box>
+        );
+    }
+
+    function renderTransactions(transactions: Transaction[]) {
+        return (
+            <Box backgroundColor="white" style={styles.transactionWrapper}>
```
domsec, line 32: Not sure background color is needed here. Could we get away with just setting a white background to the main view and handle non white backgrounds on a per component basis? Just something to note for further investigation in the future. This works fine for now.

packages/client/components/TransactionList.tsx
```diff
@@ -0,0 +1,69 @@
...
+        );
+    }
+    return <>{renderTransactions(transactions)}</>;
+}
+
+const styles = StyleSheet.create({
+    dateSeparator: {
+        fontSize: 17,
+        fontWeight: "600",
+        marginLeft: 15,
```
domsec, line 63: Should be horizontal padding of 15 and vertical padding of 5

domsec (approved):
We could definitely refactor this and make it more generic to support different list based screens planned for future features but for now this works. 

One minor thing I noticed based on the attached screenshots, make sure the transaction list is sorted by newest to oldest before any sorting occurs. Obviously nothing to do with this PR but I just wanted to mention it as a reminder.

Besides that, great work.

alexbara2000:
For the order of the transactions, it is made like this by design. This component will display the transactions in the order it sees them. This is where the sorting and filtering algorithms will come in play that is currently being designed. But yes as you mention it is important we don't forget that sorting does not occur here.

packages/client/components/TransactionList.tsx
alexbara2000: I tried it that way originally but it kept propagating the grey color of the date separations. I will look into more details when I build my screen.

PR #68, opened 2022-10-30
CRYP-76: implement frontend component for transaction details page

alexbara2000:
Added the component in charge of rendering the transaction details page. There is still some elements missing but those will be done at the same time as the changes to the API are made


alexbara2000:
image
image

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+        return `${address.substring(0, 6)}...${address.substring(address.length - 4, address.length)}`;
+    }
+
+    function getCurrencyType(): string {
+        return transaction.transactionAddress.substring(0, 2) == "0x" ? "ETH" : "BTC";
+    }
+
+    function renderHeader() {
+        return (
+            <Box style={styles.itemWrapper}>
```
jason-gerard, line 33: I like your idea here but no need to make them functions because they don't depend on any parameters.

```suggestion
    const renderHeader = {
         <Box style={styles.itemWrapper}>
```

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    return (
+        <ScrollView>
+            {renderHeader()}
+            <Box>
+                {renderBasicInfo()}
+                {renderTransactionFee()}
+                {renderOtherDetails()}
+            </Box>
```
jason-gerard, line 186: Then here you can just reference the variable containing the jsx directly, its slightly faster but in general just best practice to not use a function when its not needed. Also do you need the `Box` wrapper here? in general I think yur wrapping everything in a box but all these components already implement box so you can apply the styling to it directly, see the comments attached to the component code.

```suggestion
            {renderHeader}
            <Box>
                {renderBasicInfo}
                {renderTransactionFee}
                {renderOtherDetails}
            </Box>
```

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+                    <Box style={styles.subItemWrapper}>
+                        <HStack>
+                            <Text>Gas Limit</Text>
+                            <Text color="text.500" style={styles.gasLimit}>
+                                100 000
+                            </Text>
+                        </HStack>
+                    </Box>
+                </VStack>
+            </Box>
```
jason-gerard, line 138: VStack implements box for example so you can just move the styling to the sub component and remove half of this nesting
```suggestion
        <VStack style={styles.itemWrapper}>
            <Text style={styles.categoryTitle}>Transaction Fee</Text>
            <Box style={styles.subItemWrapper}>
                <Text color="text.500" style={styles.elementInformationTitle}>
                    Gas Price
                </Text>
                <Text>0.0000000000012323 ETH</Text>
            </Box>
            <HStack  style={styles.subItemWrapper}>
                <Text>Gas Limit</Text>
                <Text color="text.500" style={styles.gasLimit}>
                    100 000
                </Text>
            </HStack>
        </VStack>
```

yarn.lock
```diff
@@ -9378,6 +9378,7 @@ __metadata:
     eslint-config-prettier: ^8.3.0
     eslint-plugin-prettier: ^4.0.0
     eslint-plugin-unused-imports: ^2.0.0
+    expo-clipboard: ^3.1.0
```
jason-gerard, line 9381: Make sure this file get updated when you fix the package.json as well

jason-gerard (approved):
Mostly just best practice stuff so if you wanna fix it later its no problem, just make sure to fix the `package.json` and `yarn.lock` before merging because that will mess up the dependency tree fro everyone.

packages/client/components/TransactionDetails.tsx
domsec: Could even change `Box` with `VStack` with space of 40 and reduce CSS

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+        );
+    }
+    function renderBasicInfo() {
+        // TODO
+        // dynamically get the wallet name if any
+        return (
+            <Box style={styles.itemWrapper}>
+                <VStack>
+                    <Box>
```
domsec, line 60: All sub items with a title and description have a 3px spacing
```suggestion
                    <VStack space="3">
```

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+                            <Text style={styles.elementInformationText}>{transaction.transactionAddress}</Text>
+                            <Pressable onPress={() => copyToClipboard(transaction.transactionAddress)}>
+                                <FontAwesomeIcon icon={faCopyCustom} style={styles.copyIcon} size={20} />
+                            </Pressable>
+                        </HStack>
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <HStack>
+                            <Text style={styles.elementInformationText}>Status</Text>
+                            <Badge colorScheme="success" style={styles.transactionStatus}>
```
domsec, line 74: The `colorScheme` property seems to be overriding your CSS (`fontWeight`, etc). This is based on the attached screenshots. The confirmed badge designed has a color scheme so we can't simply use the built in native base success color scheme unfortunately.

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <Text color="text.500" style={styles.elementInformationTitle}>
+                            From
+                        </Text>
+                        <HStack space="10px">
+                            <Text style={styles.elementInformationText}>{transaction.walletIn}</Text>
+                            {isIncommingTransaction ? (
+                                <Pressable onPress={() => copyToClipboard(transaction.walletIn)}>
+                                    <FontAwesomeIcon icon={faCopyCustom} style={styles.copyIcon} size={20} />
```
domsec, line 87: I have no idea why from looking at your code but the copy icon (based on the screenshots provided) aligns with both the title and text instead of the text. No biggie since we can clean this up in a future PR. This comment is more for me to remember when doing the associated refactoring.

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <Text color="text.500" style={styles.elementInformationTitle}>
+                            Gas Price
+                        </Text>
+                        <Text>0.0000000000012323 ETH</Text>
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <HStack>
+                            <Text>Gas Limit</Text>
```
domsec, line 131: Should be font size of 17

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+        // Add block number, position in block and nonce
+        return (
+            <Box style={styles.itemWrapper}>
+                <VStack>
+                    <Box>
+                        <Text style={styles.categoryTitle}>Other Details</Text>
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <HStack>
+                            <Text>Block Number</Text>
```
domsec, line 152: Should be font size 17.

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+                        <HStack>
+                            <Text>Position in Block</Text>
+                            <Text color="text.500" style={styles.gasLimit}>
+                                19
+                            </Text>
+                        </HStack>
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <HStack>
+                            <Text>Nonce</Text>
```
domsec, line 168: Should be font size 17.

packages/client/components/TransactionDetails.tsx
```diff
@@ -0,0 +1,255 @@
...
+                        <HStack>
+                            <Text>Block Number</Text>
+                            <Text color="text.500" style={styles.gasLimit}>
+                                15
+                            </Text>
+                        </HStack>
+                    </Box>
+                    <Box style={styles.subItemWrapper}>
+                        <HStack>
+                            <Text>Position in Block</Text>
```
domsec, line 160: Should be font size 17.

domsec (approved):
Similar sentiment as @jason-gerard.  There is just some basic styling issues and refactoring that don't need immediate attention and could be fixed later on.

packages/client/types.tsx
```diff
@@ -25,6 +26,11 @@ export type HomeStackParamList = {
     HomeScreen: undefined;
     AddWalletSelectionScreen: undefined;
     AddWalletScreen: AddWalletScreenProps;
+    TransactionDetailsScreen: {
+        title: string;
+        transaction: Transaction;
+        walletAddress: string;
```
jason-gerard, line 32: Extract this into its own type

packages/client/screens/TransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { HomeStackScreenProps, SettingsStackScreenProps } from "../types";
+
+export default function TransactionDetailsScreen(
+    props: CompositeScreenProps<
+        HomeStackScreenProps<"TransactionDetailsScreen">,
+        SettingsStackScreenProps<"TransactionDetailsScreen">
+    >,
+) {
+    return (
+        <ScrollView style={styles.view}>
```
jason-gerard, line 15: Wrap this in a view

jason-gerard (approved):

PR #69, opened 2022-10-31
CRYP-101: auto sign out when user not found

jason-gerard:


domsec (approved):
Looks good

PR #72, opened 2022-11-01
CRYP-183: Implement UI mockup typography

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-183

## Changes
- Replace SpaceMono font with SegoeUI (regular, semibold, and bold)
- Update native base theme with the typography (font size, line height) used in the UI mockups


jason-gerard (approved):
Nice

PR #74, opened 2022-11-02
CRYP-26: Modified seed data to use wei

Andre-Ibrahim:
- updated seed data to use wei


jason-gerard (approved):
Thanks for this, will fix the system tests

PR #77, opened 2022-11-03
CRYP-130: System tests for transaction list page

alexbara2000:
I added the system tests for the transaction list page

### DIFF
--- a/packages/client/components/TransactionList.tsx
+++ b/packages/client/components/TransactionList.tsx
@@ -21,7 +21,7 @@ export function TransactionList({ transactions, walletAddress, displaySeparation
         }
         savedDate = date;
         return (
-            <Box key={+date} backgroundColor="text.100">
+            <Box key={+date} backgroundColor="text.100" testID="transactionsDateSeparator">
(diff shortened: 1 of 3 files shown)


jason-gerard (approved):
Not sure you need to test everything like that. it's pretty brittle to the exact data we have. you more want to test to make sure things are having values and correct shapes. etc. fine to start though. good try

alexbara2000:
@jason-gerard  should I change my system test to make sure the first 10 transactions of a wallet are present and then test for 3 transactions (1 for each wallet) test the actual information inside of it?

jason-gerard:
> @jason-gerard  should I change my system test to make sure the first 10 transactions of a wallet are present and then test for 3 transactions (1 for each wallet) test the actual information inside of it?

Probably would be better, but up to you, because we are making sure what we expect is there in a lower level not just that our for loop printed the right number of things if that makes sense.

alexbara2000:
Looking back on it, I think testing the transactions details in the transaction details page makes more sense. For this test I think we can leave it as is for now and refactor in future sprints.

PR #79, opened 2022-11-04
CRYP-26: refactor to single get transactions call and fix dates

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-26

### DIFF
--- a/packages/client/components/TransactionList.tsx
+++ b/packages/client/components/TransactionList.tsx
@@ -49,10 +49,9 @@ export function TransactionList({ transactions, walletAddress, displaySeparation
         return (
             <>
                 {transactions.map((transaction, i) => (
-                    <>
+                    <Box key={i}>
                         {renderHeader(new Date(transaction.createdAt.toString()))}
                         <Box
-                            key={i}
                             backgroundColor="white"
                             style={styles.transactionWrapper}
                             testID="transactionWithSepartion"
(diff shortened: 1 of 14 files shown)


packages/client/components/TransactionList.tsx
```diff
@@ -49,10 +49,9 @@ export function TransactionList({ transactions, walletAddress, displaySeparation
         return (
             <>
                 {transactions.map((transaction, i) => (
-                    <>
+                    <Box key={i}>
```
alexbara2000, line 52: What does this change?

packages/client/components/TransactionListItem.tsx
```diff
@@ -18,36 +19,28 @@ type Props = {
...
-        return `${address.substring(0, 6)}...${address.substring(address.length - 4, address.length)}`;
-    }
-
     function getCurrencyType(): string {
         return transaction.transactionAddress.substring(0, 2) == "0x" ? "ETH" : "BTC";
     }
 
-    function formatAmount(amount: string): string {
-        return amount.length > 15 ? amount.substring(0, 15) + "..." : amount;
+    function getFormattedDate(timestamp: string): string {
```
alexbara2000, line 26: Is this only used in this file?

alexbara2000 (approved):
Looks good, I am relatively impressed!

packages/client/components/TransactionList.tsx
jason-gerard: Fixes the error of each mapped item not having a unique key, you had put it on the wrong element

packages/client/components/TransactionListItem.tsx
jason-gerard: As of now, yes

PR #80, opened 2022-11-04
CRYP-101: Update wallet list when new wallet is added

jason-gerard:

## Changes
Previously when a new wallet was added it wouldnt show up on the home screen until the app was restarted because the page is cached so the api call would never happen again, and would only show through the setting page if you navigated away then back to it. Now anytime the page containing the wallets list comes into focus again we will make a new api request to get the users wallets.

### DIFF
--- a/packages/client/components/wallets-list/WalletsList.tsx
+++ b/packages/client/components/wallets-list/WalletsList.tsx
@@ -25,6 +25,9 @@ export default function WalletsList(props: Props) {
 
     React.useEffect(() => {
         (async () => {
+            // Get all the users wallets everytime the page comes in focus
+            // without this the home page (because of caching) and the settings page
+            // will not update when a new wallet is added
             if (isFocused) {
                 const wallets = await walletsGateway.findAllWallets({ id: user.id }, token);
                 setWallets(wallets);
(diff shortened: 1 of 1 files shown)


domsec (approved):
Looks good. 

Wouldn't it be better to do this call and re-cache the wallets during the add wallet processing page that way the home page loads instantly? I could see a situation where the home page is blank for a short period of time for users that have a poor internet connection. Probably something we should test in the future.

jason-gerard:
> Looks good. 
> 
> 
> 
> Wouldn't it be better to do this call and re-cache the wallets during the add wallet processing page that way the home page loads instantly? I could see a situation where the home page is blank for a short period of time for users that have a poor internet connection. Probably something we should test in the future.

I agree with you but this wasn't for any optimization just so the app would "function". I was thinking of saving all the actual optimizations caching, pre processing, etc for release 3 when we do user testing or else we won't really have much to improve.

domsec:
> > Looks good.
> > Wouldn't it be better to do this call and re-cache the wallets during the add wallet processing page that way the home page loads instantly? I could see a situation where the home page is blank for a short period of time for users that have a poor internet connection. Probably something we should test in the future.
> 
> I agree with you but this wasn't for any optimization just so the app would "function". I was thinking of saving all the actual optimizations caching, pre processing, etc for release 3 when we do user testing or else we won't really have much to improve.

Makes perfect sense those improvements to release 3. Can you make an associated JIRA ticket so we don't forget.

PR #95, opened 2022-11-09
CRYP-117: Refactor services to use insert and find instead of save

Andre-Ibrahim:
- updated to insert for user Service
- updated comments in eth-edge transaction service

### DIFF
--- a/packages/api/src/services/users.service.ts
+++ b/packages/api/src/services/users.service.ts
@@ -18,7 +18,8 @@ export class UsersService {
         }
 
         const createdUser = this.userRepository.create(signUpReq);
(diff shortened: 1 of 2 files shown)


jason-gerard (approved):
Looks good but comments shouldn't end with punctuation, I don't like the look of it. English major

Andre-Ibrahim:
> Looks good but comments shouldn't end with punctuation, I don't like the look of it. English major

this you?
image

jason-gerard:
> > Looks good but comments shouldn't end with punctuation, I don't like the look of it. English major
> 
> this you? image

Nope

PR #96, opened 2022-11-09
CRYP-205: Refactor UI components in all screens to use the custom typography

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-205

## Changes
- Refactor all components and screens to use the custom typography
- Moving forward, there is no need to add the font size, font weight and line height CSS styling to the `Text` component or any components that inherit the `Text` component (i.e. Button)
- `Text` default styling
    - `size={"body"}`
    - `fontWeight={"regular"}`
    - `color={"text.700"}`

**Old**
```typescript
<Text style={styles.textStyle}>Text</Text>

textStyle: {
    fontSize: 20,
    lineHeight: 27,
    fontWeight: "600,
}
```

**New**
```typescript
<Text size={"title3"} fontWeight={"semibold"}>Text</Text>
```


narroarrow (approved):
Looks good, well done

jason-gerard (approved):

alexbara2000:
What does "title3" mean? Are we going to have this in figma too now?

domsec:
> What does "title3" mean? Are we going to have this in figma too now?

Figma is updated with every newly added/used typography name right in the font information for text. 

So for example, in Figma when selecting any text you'll see something like "Default/Regular/Title3". Which breaks down to regular (default) font weight and size of "title3". From a font size perspective, "title3" represents font size 20. You can find all the info in terms of what each size represents either in the Components page in Figma or the typography file in the theme folder.

PR #97, opened 2022-11-09
CRYP-223: Refactor icon folder structure by font weight

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-223

## Changes
- Refactor the icon folder to follow a by font weight/style folder structure (each font style/weight has a folder)
- Added a unique prefix to each icon to easily identify which font weight it belongs to in situations where two different styles of the same icon is used on the same page
- Prefixes are as follows
    - Light: `fal`
    - Regular: `far`
    - Solid: `fas`
    - Brands: `fa`


jason-gerard (approved):
Cool, good addition this should help organize all the icons we use.

PR #98, opened 2022-11-10
CRYP-186: add btc edge create wallet

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-186

## Changes
- Create wallets and transaction module in btc-edge
- Hook up API to btc-edge
- Implement transaction backfilling in eth-edge
- Implement get wallet balance in eth-edge

## Dev notes
Adding the new transactions after the wallet is added will happen in another PR since this one is getting quite big


packages/common/src/domain/entities/transaction.ts
```diff
@@ -1,8 +1,11 @@
-import { Column, CreateDateColumn, Entity, PrimaryColumn } from "typeorm";
+import { Column, CreateDateColumn, Entity, PrimaryGeneratedColumn } from "typeorm";
 
 @Entity()
 export class Transaction {
-    @PrimaryColumn()
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column({ nullable: false })
```
alexbara2000, line 8: Why would we use id as primary number instead of transaction address?

alexbara2000 (approved):
Really good work

packages/api/test/unit/transactions/transactions.controller.spec.ts
```diff
@@ -7,6 +7,7 @@ describe("TransactionController", () => {
     let fakeTransactionService: Partial<TransactionsService>;
 
     const transaction = {
+        id: 1,
```
narroarrow, line 10: What does this do?

packages/btc-edge/src/gateways/so_chain_gateway.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Method } from "@cryptify/common/src/utils/gateway/abstract_gateway";
+import {
+    AddressBalanceResponse,
+    TransactionResponse,
+    TransactionsByWalletResponse,
+} from "@cryptify/btc-edge/src/types/so_chain_responses";
+import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
+
+@Injectable()
+export class SoChainGateway extends AbstractServiceGateway {
```
Andre-Ibrahim, line 13: Is there no SDK? if not would it be a good idea to make a refactoring task to remove the Alchemy SDK so the different Node services look the same.

Andre-Ibrahim (approved):
Looks good just one question!
is there something like the webhook we have in place for eth-edge to get the transactions once a wallet is added in chain.so . Or do we have to implement that mechanism ourselves?

jason-gerard:
> Looks good just one question! is there something like the webhook we have in place for eth-edge to get the transactions once a wallet is added in chain.so . Or do we have to implement that mechanism ourselves?

It's not a webhook they use web sockets, but i'll add that in a follow up pr

packages/btc-edge/src/gateways/so_chain_gateway.ts
jason-gerard: There is no SDK, why would we remove the Alchemy SDK though? I don't think there is a benefit of having the different edge services "look" the same. Actually one of the main parts of the edge service architecture is allowing them to be implemented completely independent from each other and decoupled from each other and they only need to provide a uniform interface.

packages/common/src/domain/entities/transaction.ts
jason-gerard: With Bitcoin a transaction can have multiple inputs and multiple outputs so the transaction address is no longer unique because our system deals with single input single output transaction

PR #99, opened 2022-11-12
CRYP-186: add new btc transactions to db

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-186

## Changes
- Add Blockchain WS integration to track new transactions
- Subscribe to new transaction messages when btc wallet is added


narroarrow (approved):
Looks good

Andre-Ibrahim (approved):
looks good if you can drop the reference to the ws that you are using that would be great.

jason-gerard:
> looks good if you can drop the reference to the ws that you are using that would be great.

Here ya go its pretty basic but works really well https://www.blockchain.com/explorer/api/api_websocket

PR #100, opened 2022-11-13
Cryp-222: create db schema for tags

alexbara2000:
Created the db schema for the tags. I also added 2 endpoints, one to get tags of a user and one to create a tag for a user


jason-gerard (commented):
I updated jira to remove that extra task you made so youll want to make a new branch + pr with the correct id, but only include the changes for the ticket its linked to, this pr contains changes for CRYP-177 and CRYP-35

PR #101, opened 2022-11-13
CRYP-208: fix transaction page rendering

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-208

## Changes
- Convert transactions list scroll view into a flat list
- Refactor transaction list component to reduce duplication and simplify some logic

## Dev notes
This pr converts the scroll view into a flat list which will only render the transactions visible at any given moment, this greatly reduces the load time from 20-30s to less than a second. Of ocurse with large list dev builds will always be slow, if you build the app and check it out you can see how fast the improvement really makes it.


narroarrow (approved):
Good fix on the bug

alexbara2000 (approved):
looks good

PR #102, opened 2022-11-13
CRYP-120: refactor address validation to library

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-120

## Changes
- Create formal function that can parse eth and btc wallets but also transaction hashes
- Use new function to clean up a bunch of code that used improper ways of figuring out the currency type from an address

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,16 @@ jobs:
               run: docker run --entrypoint yarn cryptify-btc-edge lint:check
             - name: Unit test
               run: docker run --entrypoint yarn cryptify-btc-edge workspace @cryptify/btc-edge test:unit
+    common:
+        name: Common
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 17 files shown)


domsec (approved):
Very nice. 

Few things
1. Since we won't be using a library for the address validation are we sure the regex will not backfire in the future?
2. Are there any other places where we check and/or validate the currency/address or is `TransactionDetails` the only place? I remember seeing more screens doing such checking/validation in past PRs.

jason-gerard:
> 

1. The regex ones are standard, should be fine especially since the tx addresses are not inputted by the user its just used for how the UI is rendered.

2. It is updated in the TransactionsList too, there were other places but i removed them already in a previous refactor, i shouldn't have missed anywhere else I searched for a bit

domsec:
Perfect. Good with me.

Andre-Ibrahim (approved):
Looks good!

PR #103, opened 2022-11-13
CRYP-185: implement btc edge api feature parity

Andre-Ibrahim:
- added all end point in btc-edge
- refactored transaction_service and client_service to use btc as well
- removed all static references to eth in client 
- fixed transactions being flipped(sent, received) in transaction list


packages/btc-edge/src/modules/wallets.module.ts
```diff
@@ -8,8 +8,13 @@ import { TransactionsModule } from "./transactions.module";
 import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
 
 @Module({
-    imports: [TypeOrmModule.forFeature([Wallet, Transaction]), TransactionsModule],
+    imports: [
+        TypeOrmModule.forFeature([Wallet, Transaction]),
+        TransactionsModule,
```
jason-gerard, line 13: You have the forward ref below

```suggestion
```

packages/client/components/transactions-list/TransactionListItem.tsx
```diff
@@ -17,12 +17,14 @@ type Props = {
 };
 
 export function TransactionListItem({ transaction, walletAddress, navigation }: Props) {
-    const isIncomingTransaction = walletAddress == transaction.walletIn;
+    const isIncomingTransaction = walletAddress == transaction.walletOut;
```
jason-gerard, line 20: I don't think this makes sense, an incoming transaction would be one where the `in wallet` is the same as your wallet address. Is it possible the issue is somewhere else in the code? I haven't noticed anything before

packages/client/screens/WalletOverviewScreen.tsx
```diff
@@ -52,14 +55,18 @@ export default function WalletOverviewScreen({ route, navigation }: Props) {
                             </Text>
                         </VStack>
                         <VStack>
-                            <FontAwesomeIcon icon={faEthereum} style={styles.ethereumIcon} size={40} />
+                            <FontAwesomeIcon
+                                icon={currencyIcon ? currencyIcon : faBitcoin}
```
domsec, line 59: Isn't the `faBitcoin` redundant here since the icon is being determined by the `currencyTypeToIcon` function? Can be replaced with undefined.

```suggestion
                                icon={currencyIcon ? currencyIcon : undefined}
```

domsec (approved):

packages/client/components/transactions-list/TransactionListItem.tsx
Andre-Ibrahim: do you mean in the changes I made? I haven't touched transactions is it possible that when they are added in the db during backfill that they are flipped

jason-gerard: Don't think its anything you did so maybe, but where do you see the transactions as being flipped?

Andre-Ibrahim: I just compared with my own wallet in metamask plus one of the wallet I was testing only had outgoing transactions lol

packages/client/components/TransactionDetails.tsx
```diff
@@ -42,7 +44,8 @@ export function TransactionDetails({ transaction, walletAddress }: Props) {
                     style={styles.transactionAmountInOut}
                 >
                     {isIncommingTransaction ? "+" : "-"}
-                    {getFormattedAmount(transaction.amount, CurrencyType.ETHEREUM) + " "} {getCurrencyType()}
+                    {getFormattedAmount(transaction.amount, type ? type : CurrencyType.BITCOIN) + " "}{" "}
```
domsec, line 47: Maybe dumb question but why would an if statement be needed here for the type?

jason-gerard: Its because there are type issues with how they setup the code, all this is fixed in the pr i just made

Andre-Ibrahim: since I am using a Map the return type is (CurrencyType | undefined) to make it only CurrencyType that guard is necessary, I see that Jason refactored it however so it will be removed

domsec: Okie dokie

packages/client/services/currency_service.ts
```diff
@@ -1,14 +1,29 @@
...
+        return wallets
+            .filter((wallet) => wallet.currencyType == CurrencyType.ETHEREUM)
+            .map((wallet) => Web3.utils.toBN(wallet.balance))
+            .reduce((total, balance) => total.add(balance), Web3.utils.toBN(0))
+            .toString();
+    }
+    if (type == CurrencyType.BITCOIN) {
+        return wallets
+            .filter((wallet) => wallet.currencyType == CurrencyType.BITCOIN)
+            .map((wallet) => parseFloat(wallet.balance))
```
jason-gerard, line 21: Don't worry about it for now but this is much better, will never throw error and is very fast

```suggestion
            .map((wallet) => +wallet.balance)
```

packages/client/services/currency_service.ts
```diff
@@ -1,14 +1,29 @@
...
 
-export function getWalletsTotal(wallets: WalletWithBalance[]): string {
+export function getWalletsTotal(wallets: WalletWithBalance[], type: CurrencyType): string {
     // After summing all the wallets together as a single BigNumber convert
     // the value to string, all trailing zeros are also dropped
-    return wallets
-        .map((wallet) => Web3.utils.toBN(wallet.balance))
-        .reduce((total, balance) => total.add(balance), Web3.utils.toBN(0))
-        .toString();
+    if (type == CurrencyType.ETHEREUM) {
```
jason-gerard, line 11: Always use triple equals in JS / TS so it doesnt do implicit type casting

```suggestion
    if (type === CurrencyType.ETHEREUM) {
```

jason-gerard (approved):
Good work, don't worry about the comments, more for future reference / FYI

PR #104, opened 2022-11-14
CRYP-25: remove an ethereum wallet from an account

alexbara2000:


packages/client/gateways/wallets_gateway.ts
```diff
@@ -26,4 +27,14 @@ export class WalletsGateway extends AbstractApiGateway {
 
         return this.request<WalletWithBalance[]>(Method.GET, headers, path, null);
     }
+
+    async deleteWallet(req: DeleteWalletRequest, token: string): Promise<void> {
+        console.log("help");
```
domsec, line 32: Probably want to remove this ;)

domsec (approved):

packages/api/src/controllers/wallets.controller.ts
```diff
@@ -28,4 +30,10 @@ export class WalletsController {
 
         return this.walletsService.findAll(getWalletsReq);
     }
+
+    @Delete("/users/:userId/deleteWallet")
```
jason-gerard, line 34: API path standards are important, also make sure you add the guard in

```suggestion
    @UseGuards(JwtAuthGuard, CanMutateResourceGuard)
    @Delete("/users/:userId/wallets/:walletId")
```

domsec:
Can you please also attach screenshots of the screens.

packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
```diff
@@ -30,4 +31,8 @@ export class BtcEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
     async getTransactions(_req: GetTransactionsRequest): Promise<Transaction[]> {
         return [];
     }
+
+    async deleteWallet(_req: DeleteWalletRequest): Promise<void> {
+        return;
+    }
```
jason-gerard, line 37: Youre gunna want to implement this one too

packages/api/src/services/wallets.service.ts
```diff
@@ -23,6 +24,17 @@ export class WalletsService {
         return edgeGatewayStrategy.createWallet(req);
     }
 
+    async delete(req: DeleteWalletRequest): Promise<void> {
+        if (req.currencyType != getCurrencyType(req.address)) {
+            throw new BadRequestException(ERROR_WALLET_ADDRESS_INVALID_FOR_CURRENCY(titleCase(req.currencyType)));
+        }
+
+        // Get the edge gateway strategy associated to the currency type in the request
+        // then using this gateway create the wallet in the appropriate edge service
```
jason-gerard, line 33: Please don't copy paste comments, its better to just not have them at that point

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -22,4 +25,11 @@ export class WalletsController {
         const getWalletReq = await useValidate(getWalletsSchema, params);
         return this.walletsService.findAll(getWalletReq.id);
     }
+
+    @Delete("/users/:UserId/deleteWallet")
```
jason-gerard, line 29: Update the path here, but no need for the guard in the edge service

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -22,4 +25,11 @@ export class WalletsController {
...
+    async delete(@Body() body: DeleteWalletRequest): Promise<void> {
+        body.currencyType = CurrencyType.ETHEREUM;
+        const deleteWalletReq = await useValidate(deleteWalletSchema, body);
+        return this.walletsService.delete(deleteWalletReq.address, deleteWalletReq.userId);
```
jason-gerard, line 33: Just pass the entire request obj
```suggestion
        return this.walletsService.delete(deleteWalletReq);
```

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -73,4 +73,12 @@ export class WalletsService {
             balance,
         }));
     }
+
+    async delete(address: string, userId: number): Promise<void> {
+        //Even if address is the primary key, we might want to consider going with a model
+        //where we have a primary ket composed of userId and address. This can be done in case
+        //a user decides to claim addresses that are not actually theirs.
```
jason-gerard, line 80: I'm not sure what this comment means, I'd probably delete it

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -73,4 +73,12 @@ export class WalletsService {
...
+        this.walletRepository.delete({ userId: userId, address: address });
+        this.transactionsService.delete(address);
```
jason-gerard, line 82: When calling an async function here probably want to await it
```suggestion
        await this.walletRepository.delete({ userId: userId, address: address });
        await this.transactionsService.delete(address);
```

packages/eth-edge/src/services/transactions.service.ts
```diff
@@ -58,4 +58,9 @@ export class TransactionsService {
             where: [{ walletIn: In(addresses) }, { walletOut: In(addresses) }],
         });
     }
+
+    async delete(walletAdress: string): Promise<void> {
+        this.transactionsRepository.delete({ walletIn: walletAdress });
+        this.transactionsRepository.delete({ walletOut: walletAdress });
```
jason-gerard, line 64: Should be able to do this in 1 line as well, doesn't need to be 2 separate calls
```suggestion
        await this.transactionsRepository.delete({ walletIn: walletAdress });
        await this.transactionsRepository.delete({ walletOut: walletAdress });
```

packages/client/screens/WalletSettingsScreen.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    function handleDeleteWallet(): void {
+        Alert.alert("Do you want to remove this wallet?", "You cannot undo this action.", [
+            { text: "Cancel", style: "cancel" },
+            {
+                text: "Remove",
+                style: "destructive",
+                onPress: () => {
+                    walletsGateway.deleteWallet(
+                        { userId: user.id, address: address, currencyType: getCurrencyType(address) },
```
jason-gerard, line 25: CurrencyType is passed as a prop here

```suggestion
                        { userId: user.id, address: address, currencyType },
```

packages/client/types.tsx
```diff
@@ -56,6 +56,16 @@ type TransactionsListScreenProps = {
     displaySeparation: boolean;
 };
 
+type ViewWalletsScreenProps = {
+    isSettingsTab: boolean;
```
jason-gerard, line 60: Props drilling is a pretty bad idea and ends up in really messy code, I sent those docs that can tell you which stack you are on, it didn't work?

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -22,4 +25,11 @@ export class WalletsController {
         const getWalletReq = await useValidate(getWalletsSchema, params);
         return this.walletsService.findAll(getWalletReq.id);
     }
+
+    @Delete("/users/:UserId/deleteWallet")
+    async delete(@Body() body: DeleteWalletRequest): Promise<void> {
+        body.currencyType = CurrencyType.ETHEREUM;
```
jason-gerard, line 31: Not sure whats going on here

packages/client/screens/WalletSettingsScreen.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const walletsGateway = new WalletsGateway();
+
+    function handleDeleteWallet(): void {
+        Alert.alert("Do you want to remove this wallet?", "You cannot undo this action.", [
+            { text: "Cancel", style: "cancel" },
+            {
+                text: "Remove",
+                style: "destructive",
+                onPress: () => {
+                    walletsGateway.deleteWallet(
```
jason-gerard, line 24: This should be awaited as well or else the wallet might still show up after if it wasn't deleted in time

packages/client/screens/WalletSettingsScreen.tsx
alexbara2000: currencyType was retiring the value "ETH" which was not recognized as a currency type in the api. This solves the issue.

packages/eth-edge/src/controllers/wallets.controller.ts
alexbara2000: Me neither

alexbara2000: This was because I passed a string as the currency type instead of the the actual currency type. From what I understand master is now fixed and should work properly now

packages/client/types.tsx
alexbara2000: The docs talk about where you are on in a specific stack and you can get the root of a specific stack. We don't have that, we have a composite stack which is not supported. I could not find anything that works with composite stack and getting the proper path.

packages/client/screens/WalletSettingsScreen.tsx
jason-gerard: Yep because you guys had mixed up the currency type and the iso code, I already fixed it you just gotta rebase then currencyType prop will give the correct value

packages/client/types.tsx
jason-gerard: This can still be done by just using the route name, if this is the only time we need to do this i guess its fine but if its required again we should do it the correct way because this is really messy

packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
alexbara2000: The ticket is for Beth only but yes in the near future we can also implement it for btw

jason-gerard: its the same thing the ticket should be for both

packages/eth-edge/src/services/wallets.service.ts
alexbara2000: I thought the db was setup wrong but turns out I was wrong

packages/eth-edge/src/services/transactions.service.ts
alexbara2000: I wanted to ass a where clause but from what I can tell delete does not support it. The solution is using query builder which I don't think is a good option

alexbara2000: Can you let me know if you find anything on where clauses for delete?

packages/client/types.tsx
alexbara2000: I can't figure it out. I added a todo and I can look at it when I have more time.

packages/api/src/controllers/wallets.controller.ts
```diff
@@ -28,4 +31,11 @@ export class WalletsController {
 
         return this.walletsService.findAll(getWalletsReq);
     }
+
+    @UseGuards(JwtAuthGuard, CanMutateResourceGuard)
+    @Delete("/users/:id/wallets/:address")
+    async delete(@Body() body: DeleteWalletRequest): Promise<Wallet> {
```
jason-gerard, line 37: So DELETE requests don't have any body, this should just be taken from the parameters

packages/api/src/gateways/edge-gateway/eth_edge_gateway_strategy.ts
```diff
@@ -30,4 +32,9 @@ export class EthEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
         const path = `users/${req.id}/transactions`;
         return this.request<Transaction[]>(Method.GET, {}, path, null);
     }
+
+    async deleteWallet(req: DeleteWalletRequest): Promise<Wallet> {
+        const path = `users/${req.userId}/wallets/${req.address}`;
+        return this.request<Wallet>(Method.DELETE, {}, path, req);
```
jason-gerard, line 38:
```suggestion
        return this.request<Wallet>(Method.DELETE, {}, path, null);
```

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -73,4 +75,14 @@ export class WalletsService {
             balance,
         }));
     }
+
+    async delete(deleteWalletReq: DeleteWalletRequest): Promise<Wallet> {
+        const walletToDelete = await this.findOne(deleteWalletReq.address, deleteWalletReq.userId);
+        if (!walletToDelete) {
+            throw new BadRequestException(ERROR_WALLET_NAME_DOES_NOT_EXIST);
+        }
```
jason-gerard, line 83: Not sure thers any benefit of checking to see if it exists first
```suggestion
```

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -73,4 +75,14 @@ export class WalletsService {
...
+        await this.walletRepository.delete({ address: deleteWalletReq.address, userId: deleteWalletReq.userId });
+        await this.transactionsService.delete(deleteWalletReq.address);
```
jason-gerard, line 85: This isn't being done correctly because you aren't taking into account that other users can have this wallet tracked or that other users are using wallets invlolved in transactions with this wallet, as it is now all those transactions will be improperly removed.

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -73,4 +75,14 @@ export class WalletsService {
...
+        return walletToDelete;
```
jason-gerard, line 86: Its good to return the wallet here but should be the wallet with a balance not just wallet

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -22,4 +25,11 @@ export class WalletsController {
         const getWalletReq = await useValidate(getWalletsSchema, params);
         return this.walletsService.findAll(getWalletReq.id);
     }
+
+    @Delete("/users/:UserId/wallets/:address")
+    async delete(@Body() body: DeleteWalletRequest): Promise<Wallet> {
+        const deleteWalletReq = await useValidate(deleteWalletSchema, body);
+        console.log(deleteWalletReq);
```
jason-gerard, line 32: Remove extra log

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -22,4 +25,11 @@ export class WalletsController {
...
+        return await this.walletsService.delete(deleteWalletReq);
```
jason-gerard, line 33: No need to await a return statement

```suggestion
        return this.walletsService.delete(deleteWalletReq);
```

packages/eth-edge/src/controllers/wallets.controller.ts
```diff
@@ -22,4 +25,11 @@ export class WalletsController {
         const getWalletReq = await useValidate(getWalletsSchema, params);
         return this.walletsService.findAll(getWalletReq.id);
     }
+
+    @Delete("/users/:UserId/wallets/:address")
+    async delete(@Body() body: DeleteWalletRequest): Promise<Wallet> {
```
jason-gerard, line 30: Same thing as above, DELETE shouldn't have any body

packages/common/src/validations/delete_wallet_schema.ts
```diff
@@ -0,0 +1,8 @@
+import * as yup from "yup";
+import { CurrencyType } from "@cryptify/common/src/domain/currency_type";
+
+export const deleteWalletSchema = yup.object({
+    userId: yup.number().required(),
+    address: yup.string().required("Enter a wallet address."),
+    currencyType: yup.mixed<CurrencyType>().oneOf(Object.values(CurrencyType)).required(),
```
jason-gerard, line 7: No body for this request and we can extract the currency type anyways, so we can remove this

```suggestion
```

packages/common/src/utils/gateway/abstract_gateway.ts
```diff
@@ -11,7 +11,6 @@ export abstract class AbstractGateway {
             },
             body: body ? JSON.stringify(body) : null,
         });
-
```
jason-gerard, line 14: ?

packages/common/src/errors/error_messages.ts
```diff
@@ -6,5 +6,6 @@ export const ERROR_WALLET_ADDRESS_INVALID_FOR_CURRENCY = (currency: string) =>
 export const ERROR_WALLET_ALREADY_ADDED_TO_ACCOUNT = (currency: string) =>
     `address:This wallet address is already associated with another ${currency} wallet.`;
 export const ERROR_WALLET_NAME_ALREADY_ADDED_TO_ACCOUNT = "name:This name is already associated with another wallet.";
+export const ERROR_WALLET_NAME_DOES_NOT_EXIST = "Wallet does not Exist";
```
jason-gerard, line 9: Don't think we need this error message for anything

packages/client/gateways/wallets_gateway.ts
```diff
@@ -26,4 +27,13 @@ export class WalletsGateway extends AbstractApiGateway {
...
         return this.request<WalletWithBalance[]>(Method.GET, headers, path, null);
     }
+
+    async deleteWallet(req: DeleteWalletRequest, token: string): Promise<void> {
+        const path = `users/${req.userId}/wallets/${req.address}`;
+        const headers = {
+            Authorization: `Bearer ${token}`,
+        };
+
+        return this.request<void>(Method.DELETE, headers, path, req);
```
jason-gerard, line 37:
```suggestion
        return this.request<void>(Method.DELETE, headers, path, null);
```

packages/api/src/services/wallets.service.ts
```diff
@@ -23,6 +25,11 @@ export class WalletsService {
         return edgeGatewayStrategy.createWallet(req);
     }
 
+    async delete(req: DeleteWalletRequest): Promise<Wallet> {
+        const edgeGatewayStrategy = this.edgeGatewayStrategyFactory.get(req.currencyType);
```
jason-gerard, line 29: Here you can extract the currency type from the address, once the body param is removed

packages/api/src/gateways/edge-gateway/abstract_edge_gateway_strategy.ts
```diff
@@ -4,13 +4,16 @@ import { WalletWithBalance } from "@cryptify/common/src/domain/wallet_with_balan
...
+import { DeleteWalletRequest } from "@cryptify/common/src/requests/delete_wallet_request";
+import { Wallet } from "@cryptify/common/src/domain/entities/wallet";
 
 export abstract class AbstractEdgeGatewayStrategy extends AbstractServiceGateway {
     protected constructor(uri: string) {
         super(uri);
     }
 
     abstract createWallet(req: CreateWalletRequest): Promise<WalletWithBalance>;
+    abstract deleteWallet(req: DeleteWalletRequest): Promise<Wallet>;
```
jason-gerard, line 16: Again if youre going to return the wallet it should be with the balance

```suggestion
    abstract deleteWallet(req: DeleteWalletRequest): Promise<WalletWithBalance>;
```

packages/api/src/controllers/wallets.controller.ts
```diff
@@ -28,4 +31,11 @@ export class WalletsController {
...
+        const deleteWalletReq = await useValidate(deleteWalletSchema, body);
+        return await this.walletsService.delete(deleteWalletReq);
```
jason-gerard, line 39:
```suggestion
        return this.walletsService.delete(deleteWalletReq);
```

packages/api/src/controllers/wallets.controller.ts
alexbara2000: @jason-gerard  "/users/:id/wallets/:address/currencyType/:currencyType" ?

packages/eth-edge/src/services/wallets.service.ts
alexbara2000: So what is the solution?

packages/eth-edge/src/services/wallets.service.ts
alexbara2000: Why?

jason-gerard: Because that's what the frontend uses. look at all the other return types

packages/api/src/controllers/wallets.controller.ts
jason-gerard: look below. i don't think you need to pass that at all

alexbara2000:
image
image

packages/eth-edge/src/services/wallets.service.ts
alexbara2000: @jason-gerard I propose we don't delete any transactions if a wallet is deleted for now. are you ok with that?

packages/api/src/guards/can_mutate_resource.guard.ts
```diff
@@ -9,16 +9,14 @@ export class CanMutateResourceGuard implements CanActivate {
      * @param context
      */
     canActivate(context: ExecutionContext): boolean | Promise<boolean> | Observable<boolean> {
-        const { body, params, user } = context.switchToHttp().getRequest();
-
-        if (!body.userId) {
+        const { body, params, user, method } = context.switchToHttp().getRequest();
```
jason-gerard, line 12: Stop removing all the spacing

packages/common/src/utils/gateway/abstract_gateway.ts
jason-gerard: Stop removing my spacing

packages/eth-edge/src/controllers/wallets.controller.ts
jason-gerard: fix this

packages/eth-edge/src/services/wallets.service.ts
jason-gerard: Yea so for now just delete this line and the transaction service method

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -73,4 +74,21 @@ export class WalletsService {
...
+        //That means we cannot pass it to the delete method from transactionService
+        const deepCopy = new Wallet();
+        deepCopy.address = walletToDelete.address;
+        deepCopy.userId = walletToDelete.userId;
+        deepCopy.name = walletToDelete.name;
+        deepCopy.currencyType = walletToDelete.currencyType;
+
+        await this.walletRepository.remove(walletToDelete);
+        await this.transactionsService.delete(deepCopy);
+        return { ...deepCopy, balance };
```
jason-gerard, line 92:
```suggestion
        const wallet = await this.findOne(deleteWalletReq.address, deleteWalletReq.id);

        const [balance, count] = await Promise.all([
            this.alchemyNodeServiceFacade.getBalance(deleteWalletReq.address),
            this.walletRepository.countBy({ address: deleteWalletReq.address });
        ]);
        
        await this.walletRepository.delete({ address: deleteWalletReq.address, userId: deleteWalletReq.id });
        
        if (count == 1) {
            // If count is 1 then this was the only user who had this wallet and so we can proceed with the transaction clean up,
            // the cleanup process can also be done asynchronously because we don't have to worry about UI issue because no users
            // will see those transactions anyways
            this.transactionsService.cleanup(deleteWalletReq.address);
        }

        return { ...wallet, balance };
```

jason-gerard (commented):
Fix up those 2 methods like I showed and the other comments I left that aren't resolved, then it should be good

packages/eth-edge/src/services/transactions.service.ts
```diff
@@ -58,4 +59,20 @@ export class TransactionsService {
...
+        const transactionToDelete = await this.transactionsRepository.query(
+            `select * from transaction as t
+        where t."walletOut" = $1
+        and t."walletIn" not in (select lower(address) from wallet)
+        or t."walletIn" = $1
+        and t."walletOut" not in (select lower(address) from wallet)`,
+            [wallet.address.toLowerCase()],
+        );
+        return await this.transactionsRepository.remove(transactionToDelete);
+    }
```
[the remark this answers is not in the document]
alexbara2000: @jason-gerard I see you deleted the await. Is there any reason? I was thinking that transaction cleanup we could do it async. That way the user is not stuck waiting for for transactions to be cleanup when they press remove wallets. The removing of wallet will be done sync and transactions cleanup async. Does that make sense?

jason-gerard (approved):
Looks good

packages/eth-edge/src/services/wallets.service.ts
jason-gerard: Works with me

PR #105, opened 2022-11-14
CRYP-226: fix transaction in and out flipped

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-226

## Changes
- Fix ETH transactions being stored backwards
- Fix BTC transactions being stored backwards
- Implement BTC reverse transaction pooling to improve accuracy of MIMO transaction amounts

### DIFF
--- a/packages/btc-edge/src/gateways/so_chain_gateway.ts
+++ b/packages/btc-edge/src/gateways/so_chain_gateway.ts
@@ -4,7 +4,10 @@ import { AbstractServiceGateway } from "@cryptify/common/src/utils/gateway/abstr
 import { Method } from "@cryptify/common/src/utils/gateway/abstract_gateway";
 import {
     AddressBalanceResponse,
+    PairsWithAmount,
     TransactionResponse,
+    TransactionResponseInput,
+    TransactionResponseOutput,
     TransactionsByWalletResponse,
 } from "@cryptify/btc-edge/src/types/so_chain_responses";
 import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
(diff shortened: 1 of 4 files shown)


packages/btc-edge/src/gateways/so_chain_gateway.ts
```diff
@@ -31,53 +34,62 @@ export class SoChainGateway extends AbstractServiceGateway {
...
         const txData = await this.request<TransactionResponse>(Method.GET, {}, path, null);
 
-        // Cross the in and out wallets to form NxM number of transactions then construct the
-        // transaction object, note: wallet in amounts + fee = wallet out amount
-        return txData.data.inputs
-            .flatMap((input) => txData.data.outputs.map((output) => [input, output] as const))
-            .map(([inWallet, outWallet]) =>
+        // Reverse the pool transaction to get NxM pairs and then construct the transaction entity from each
+        // pair + amount tuple
+        // Note: wallet out amounts + fee = wallet in amount
```
domsec, line 63: Doesn't the wallet in not include the fee?

domsec (approved):
Looks good

packages/btc-edge/src/gateways/so_chain_gateway.ts
jason-gerard: I left a comment lower in the code but its not intuitive which is in and out, inwallets are inputting into the transaction pool, where the out wallets take from and where the fee is taken from. so in = out + fee

domsec: I see. I believe the api sent amount value includes the fee from what I remember. So lets just make sure we aren't including the fee twice.

jason-gerard: The endpoint we are using doesnt include the total "sent amount" so I'm not sure what that includes but for example if you take this transaction

https://sochain.com/api/v2/get_tx/BTC/57b7ffb1cc862488129527bffec8048fab7abfb4aae67c409767b1e142c83c64

You can sum the 2 outputs + fee and it will equal the single input value

domsec: I was indeed looking at a different endpoint that included the fee in the amount. This PR can be merged.

Andre-Ibrahim (approved):
Nice, I reverted the changes I had in the client side to see the transactions properly.

PR #106, opened 2022-11-14
CRYP-120: fix wallet validation lib for react native and m1

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-120


packages/common/src/utils/wav.ts
```diff
@@ -0,0 +1,6020 @@
...
+    var codePoint = null
+    var bytesPerSequence = (firstByte > 0xEF) ? 4
+      : (firstByte > 0xDF) ? 3
+        : (firstByte > 0xBF) ? 2
+          : 1
+
+    if (i + bytesPerSequence <= end) {
+      var secondByte, thirdByte, fourthByte, tempCodePoint
+
+      switch (bytesPerSequence) {
```
Andre-Ibrahim, line 1108: switch case?

Andre-Ibrahim (approved):
looks good hopefully it works with the other m1 dependency issues in the future!

packages/common/src/utils/wav.ts
jason-gerard: Yes this is a programming primitive keyword found in most if not all modern general purpose proramming languages. The main goal of a switch case is to avoid the implicit order an if/elif/else tree makes, ensure all the conditionals are based on the same values, and even to allow fall through in cases where a single truth means multiple actions.

Andre-Ibrahim: good to know, will use it in my next pr!!

PR #108, opened 2022-11-15
CRYP-000: Skip failing test

jason-gerard:
## Context
There seems to be some issue with the websocket service starting up in the tests before the repository has made a connection with the DB. This issue doesn't appear when the application is actually running in dev or prod. For the time being the test can be skipped so it doesn't block others


alexbara2000 (approved):
looks good

PR #109, opened 2022-11-15
CRYP-171: create domain model

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-171


domsec (approved):
Looks good. 

Few things:
1. Should we have `ProUser` in the domain model right now given nothing has been determined with it from both a design and dev perspective? What about a basic user?
2. More of a personal preference but I'm not a fan of `WalletTag` given that for a wallet it is simply a custom name assigned to a wallet and for transaction its a way to group and classify many transactions together. So we're using the word `tag` interchangeably to mean two very different things.

jason-gerard:
> Looks good. 
> 
> 
> 
> Few things:
> 
> 1. Should we have `ProUser` in the domain model right now given nothing has been determined with it from both a design and dev perspective? What about a basic user?
> 
> 2. More of a personal preference but I'm not a fan of `WalletTag` given that for a wallet it is simply a custom name assigned to a wallet and for transaction its a way to group and classify many transactions together. So we're using the word `tag` interchangeably to mean two very different things.

1. Just from the fact we have it with our personas we should include it. Domain model should just be from a high level view so doesn't matter too much if it's not implemented. Ill add the basic user since that's a persona too. 
2. That's just how they have been referred to so far what could be better names? Contact for wallet and Category for transaction?

domsec:
> > Looks good.
> > Few things:
> > 
> > 1. Should we have `ProUser` in the domain model right now given nothing has been determined with it from both a design and dev perspective? What about a basic user?
> > 2. More of a personal preference but I'm not a fan of `WalletTag` given that for a wallet it is simply a custom name assigned to a wallet and for transaction its a way to group and classify many transactions together. So we're using the word `tag` interchangeably to mean two very different things.
> 
> 1. Just from the fact we have it with our personas we should include it. Domain model should just be from a high level view so doesn't matter too much if it's not implemented. Ill add the basic user since that's a persona too.
> 2. That's just how they have been referred to so far what could be better names? Contact for wallet and Category for transaction?

1. We don't have basic and pro personas. Basic and Pro will eventually designate account type. It would probably be more like Beginner, Intermediate, Expert ISA Person and Person has a UserAccount something like that. But like you said Domain Model is a high level view so not sure we want to go that detailed.
2. Contact should be a separate thing (User has many Contacts, Contact is a User, etc) and Category once designed/implemented should be its own thing with TransactionCategory being connect to both a Transaction and Category. To be honest, I'm not entirely sure we even need WalletTag just going based on the ERD. Wallet tagging (i.e. setting a custom name) is a requirement in the add wallet process and since there's a composite key in Wallet (wallet address and user id) and a name attribute as well why even need Wallet tag if it contains the same info as Wallet?

Now that I've had another look, there should also be a separate Tag that connects to User/UserAccount as well since tags are account wide and TransactionTag contains tags from Tag.

I'm open to also just discussing this on Discord as well.

Andre-Ibrahim (approved):

alexbara2000 (approved):
Wow, amazing coding

PR #110, opened 2022-11-15
CRYP-170: create use case diagram

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-170


domsec (commented):
At a first glance, it's missing sign up, sign in, sign out.  There's probably a few more missing that I'm forgetting about right now but this is a good start.

On a side note, how detailed are the use case diagram expected to be? Should we have 1 use case for main feature?

jason-gerard:
> At a first glance, it's missing sign up, sign in, sign out. There's probably a few more missing that I'm forgetting about right now but this is a good start.
> 
> On a side note, how detailed are the use case diagram expected to be? Should we have 1 use case for main feature?

SIgn in, sign up, and sign out aren't use cases they are costraints on the system i.e. non functional requirements. The use cases are like what the user can do with the system. No one "wants" to sign up or sign in, they want to check their wallets, but they have to sign in to do that.

There are much better explanations here https://softwareengineering.stackexchange.com/questions/370739/when-log-in-should-be-a-use-case-and-when-not

In general I think use case should be high level enough that there isn't clutter, like I don't think having a specific use case for "remove wallet" would make sense that's why I bundle them up together.

Andre-Ibrahim (approved):
looks good

alexbara2000 (approved):
Looks decent

PR #111, opened 2022-11-16
CRYP-177: Created the db schema for the tag table.

alexbara2000:
I added the tag entity and all dependencies.


packages/common/src/domain/entities/tag.ts
```diff
@@ -0,0 +1,20 @@
+import { Entity, JoinColumn, JoinTable, ManyToMany, ManyToOne, PrimaryColumn } from "typeorm";
+import { Transaction } from "./transaction";
+ import { User } from "./user";
+
+ @Entity()
+ export class Tag {
```
jason-gerard, line 6: To be more specific we should probably prefix the tag with transaction?

```suggestion
 export class TransactionTag {
```

packages/common/src/domain/entities/tag.ts
```diff
@@ -0,0 +1,20 @@
...
+     @PrimaryColumn()
+     tagName: string;
+
+     @PrimaryColumn()
+     userId: Number;
```
jason-gerard, line 11: You don't need to inlcude this here you can reference the user id from the attribute for the primary key

jason-gerard: I'm not sure if you are even able to do this though so maybe leave

packages/common/src/domain/entities/TransactionTag.ts
```diff
@@ -0,0 +1,20 @@
...
+    @PrimaryColumn()
+    userId: number;
+
+    @ManyToOne(() => User, (user) => user.tags)
+    @JoinColumn({ name: "userId", referencedColumnName: "id" })
+    user: User;
+
+    @ManyToMany(() => Transaction)
+    @JoinTable()
+    Transactions: Transaction[];
```
jason-gerard, line 19: Lowercase t "transactions"

packages/common/src/domain/entities/user.ts
```diff
@@ -23,4 +24,7 @@ export class User {
 
     @OneToMany(() => Wallet, (wallet) => wallet.user)
     wallets: Wallet[];
+
+    @OneToMany(() => TransactionTag, (transactionTag) => transactionTag.user)
+    tags: TransactionTag[];
```
jason-gerard, line 29: transactionTags

jason-gerard (approved):
Just some naming stuff you should be able to fix quick

Andre-Ibrahim (approved):
Looks good, great job!

PR #115, opened 2022-11-19
CRYP-35: Created API for create a transaction

alexbara2000:


packages/api/src/services/tags.service.ts
```diff
@@ -13,4 +15,13 @@ export class TagsService {
...
         return this.tagRepository.findBy({ userId });
     }
+
+    async create(req: CreateTagRequest): Promise<TransactionTag> {
+        if (await this.tagRepository.findOneBy(req)) {
+            throw new BadRequestException(ERROR_TAG_NAME_ALREADY_ADDED_TO_ACCOUNT);
+        }
+        const createdTag = this.tagRepository.create(req);
+        await this.tagRepository.insert(createdTag);
+        return await this.tagRepository.findOneBy(createdTag);
```
Andre-Ibrahim, line 25: two calls to findOneBy...not very optimal

Andre-Ibrahim (approved):
looks good

packages/common/src/validations/create_tag_schema.ts
```diff
@@ -0,0 +1,6 @@
+import * as yup from "yup";
+
+export const createTagSchema = yup.object({
+    userId: yup.number().required(),
+    tagName: yup.string().required(),
```
Andre-Ibrahim, line 5: shoudln't there be a list of transactions here]

alexbara2000: https://www.figma.com/file/Igge56wzf6UAWZwFcCzglv/Handoff?node-id=461%3A25941&t=GD6mc0CNpdeawjVI-0
If you take a look at the design specs you could see that you can create tags from the settings page and they are not related to any transactions. I don't know if in the future you can create a tag directly from the transaction but as of now I just followed the ticket requirement.

packages/api/src/services/tags.service.ts
alexbara2000: micro optimization for non native cloud based application can actually reduce performance and hurt the general metrics of the project. This is done a confirmation of the relevant information sent by the database to the api.

packages/api/src/services/tags.service.ts
```diff
@@ -13,4 +15,13 @@ export class TagsService {
     async findAll(userId: number): Promise<TransactionTag[]> {
         return this.tagRepository.findBy({ userId });
     }
+
+    async create(req: CreateTagRequest): Promise<TransactionTag> {
+        if (await this.tagRepository.findOneBy(req)) {
+            throw new BadRequestException(ERROR_TAG_NAME_ALREADY_ADDED_TO_ACCOUNT);
+        }
+        const createdTag = this.tagRepository.create(req);
```
jason-gerard, line 23: Do you like this name better? because I do

```suggestion
        const tag = this.tagRepository.create(req);
```

packages/api/src/services/tags.service.ts
jason-gerard, line 25: return === no await, its convention to not await the return when we want to return a promise. This can sometimes change inside a try catch but thats for a different time

```suggestion
        return this.tagRepository.findOneBy(tag);
```

jason-gerard (approved):
Fantastico!

PR #117, opened 2022-11-21
CRYP-35: Create new tags from a screen

alexbara2000:
IMG_2184
IMG_2183
IMG_2182
@domsec If we use the return as the way to add the tag, it will not work properly. The error validation does not work. The form can't be reset in a proper way. The keyboard will disappear when you press return. That is why I think a button might be a better solution. Let me know what you think of it. With the merge deadline coming soon, I don't know how much more time I can spend on the return option.


domsec:
@alexbara2000 There's definitely a way to get the return to work the way we want it to. Let me have a look at it and I'll see what I can do. If a solution can't be found then the button will have to suffice for now until we find a solution that follows the specs in a future sprint. Give me a day to look into this.

packages/client/screens/AddTagsScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+                )}
+            </Formik>
+            {newTags.length != 0 && 
+                <>
+                    <Text fontWeight={"semibold"} size={"title3"} style={styles.tagList}>
+                        Tags Added
+                    </Text>
+
+                    <HStack flexWrap="wrap" space="13">
+                        {newTags.map((tag, i) => (
```
domsec, line 72: The tags in the Tags Added section need to be ordered in alphabetical order by tag name

packages/client/screens/AddTagsScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+                {({ values, handleChange, errors, touched, submitForm }) => (
+                    <HStack space="5">
+                        <FormControl isInvalid={!!(errors.tag && touched.tag)} style={styles.tagForm}>
+                            <Input
+                                value={values.tag}
+                                onChangeText={handleChange("tag")}
+                                placeholder="For e.g., Trip"
+                                testID="AddedtagName"
+                                maxLength={20}
+                            />
```
domsec, line 56: By adding the following props we're able to get the form to submit through the keyboard and the keyboard to maintain focus as specified in the specs. Thus we no longer need the add a tag form button

Props added:
- `onSubmitEditing` submits the form when the `Return` keyboard key is pressed
- `enablesReturnKeyAutomatically` disables the `Return` keyboard key when the input field is empty and enables it when there is at least 1 character in the input. 
- `keybardType` sets the device keyboard to be in ASCII mode since we only want ASCII support for the tag name.
- `autoFocus` automatically focuses the input field on page load
- `ref` provides a way to reference the component. 
    - `{(input) => input?.focus()}` is what handles the automatic refocus of the input tag after each form submit

```suggestion
                            <Input
                                value={values.tag}
                                onChangeText={handleChange("tag")}
                                placeholder="For e.g., Trip"
                                testID="AddedtagName"
                                maxLength={20}
                                keyboardType={"ascii-capable"}
                                onSubmitEditing={submitForm}
                                enablesReturnKeyAutomatically={true}
                                autoFocus={true}
                                ref={(input) => input?.focus()}
                            />
```

packages/client/screens/AddTagsScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+                            <FormControl.ErrorMessage>{errors.tag}</FormControl.ErrorMessage>
+                        </FormControl>
+                        <Button onPress={submitForm} style={styles.addTagButton}>
+                            Add a tag
+                        </Button>
+                    </HStack>
+                )}
+            </Formik>
+            {newTags.length != 0 &&
```
domsec, line 65: When enough tags are added, the Tags Added section is not scrollable to see tags hidden by the keyboard.

packages/client/screens/AddTagsScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+                                onChangeText={handleChange("tag")}
+                                placeholder="For e.g., Trip"
+                                testID="AddedtagName"
+                                maxLength={20}
+                            />
+                            <FormControl.ErrorMessage>{errors.tag}</FormControl.ErrorMessage>
+                        </FormControl>
+                        <Button onPress={submitForm} style={styles.addTagButton}>
+                            Add a tag
+                        </Button>
```
domsec, line 61: Remove and any associated styling now that a solution has been found.

domsec (approved):
@alexbara2000 I have provided the code needed to get the form to follow the specs outlined in the mockups.

PR #118, opened 2022-11-22
CRYP 178: edit user tag frontend

Andre-Ibrahim:

### DIFF
--- a/packages/client/screens/EditTagScreen.tsx
+++ b/packages/client/screens/EditTagScreen.tsx
@@ -11,11 +11,15 @@ export default function EditTagScreen({ route }: SettingsStackScreenProps<"EditT
     const tagsGateway = new TagsGateway();
     const { token, user } = React.useContext(AuthContext);
 
-    const [currentTagName, SetCurrentTagName] = React.useState<string>(route.params.tag.tagName);
+    const [currentTagName, setCurrentTagName] = React.useState<string>(route.params.tag.tagName);
 
     const toast = useToast();
 
-    async function onSubmitEditTag(values: any, formikHelpers: FormikHelpers<any>) {
+    type TagValue = {
+        tag: string,
+    }
+
+    async function onSubmitEditTag(values: TagValue, formikHelpers: FormikHelpers<TagValue>) {
         try {
             const updatedTag = await tagsGateway.updateTag(
                 {
@@ -26,8 +30,9 @@ export default function EditTagScreen({ route }: SettingsStackScreenProps<"EditT
                 token,
             );
 
-            SetCurrentTagName(updatedTag.tagName);
-
+            setCurrentTagName(updatedTag.tagName);
+            
+            //TO DO refactor this into a separate toast component in the future.
             toast.show({
                 placement: "top",
                 duration: 2000,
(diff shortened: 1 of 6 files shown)


narroarrow (approved):
Looks good

packages/client/screens/EditTagScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { StyleSheet } from "react-native";
+import { Formik, FormikHelpers } from "formik";
+import { TagsGateway } from "../gateways/tags_gateway";
+import { AuthContext } from "../components/contexts/AuthContext";
+
+export default function EditTagScreen({ route }: SettingsStackScreenProps<"EditTagScreen">) {
+    const tagsGateway = new TagsGateway();
+    const { token, user } = React.useContext(AuthContext);
+
+    const [currentTagName, SetCurrentTagName] = React.useState<string>(route.params.tag.tagName);
```
jason-gerard, line 14: Camel case...

packages/client/screens/EditTagScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    const toast = useToast();
+
+    async function onSubmitEditTag(values: any, formikHelpers: FormikHelpers<any>) {
```
jason-gerard, line 18: Add types here

packages/client/screens/TagsSettingsScreen.tsx
```diff
@@ -35,15 +67,24 @@ export default function TagsSettingsScreen() {
             <HStack flexWrap="wrap" space="13">
                 {tags.map((tag, i) => (
                     // TODO: Create a custom badge component. More information can be found https://github.com/SOEN-490-Capstone/Cryptify/pull/113
-                    <Badge
```
jason-gerard, line 38: Delete the comment too if we aren't using the badge anymore?

packages/client/screens/TagsSettingsScreen.tsx
```diff
@@ -21,6 +23,36 @@ export default function TagsSettingsScreen() {
             setTags(tags);
         })();
     });
+    const [isEdit, setIsEdit] = React.useState<boolean>(false);
```
jason-gerard, line 26: isEditMode maybe idk

jason-gerard (approved):
Coo

domsec (approved):
Looks good

packages/client/screens/TagsSettingsScreen.tsx
Andre-Ibrahim: there is no benefit of using this badge component when we want it completely different from the base one anyway imo.

domsec: The main benefit is the ability to utilize the default native base theme for the badge component to ensure it's uniform across all screens. Adding a icon and making it pressable shouldn't need a completely custom badge implementation since at the core it is still a badge.

Andre-Ibrahim: Maybe we can create a custom Badge component but I would still recommend to not use the one Native-Base provides since there are some limitations for such basic UI component.

domsec: What kind of limitations? I haven't looked that deep into the badge component nor its theming so any information would be a great help to clarify what I'm clearly not understanding about why we would need to go the custom route.

Andre-Ibrahim: can't really tell you exactly but some weird behavior with the padding, also if you look at the code changes I made the only change in the props is the name color for background color, not sure if the Badge does anything here other than abstractions we might not even know about.

domsec: Your guess is as good as mine. I'll have to take a look at it in the future for further clarification.

PR #119, opened 2022-11-23
CRYP-25: Added ability to delete a BTC wallet

alexbara2000:


packages/btc-edge/src/services/transactions_service.ts
```diff
@@ -31,4 +31,21 @@ export class TransactionsService {
...
+
+    async cleanup(address: string): Promise<Transaction[]> {
+        // Here we get all the transactions that involve the wallet we are removing but that
+        // don't involve wallets still registered in our system
+        // TODO find a faster way to execute this query
+        const transactions = await this.transactionsRepository.query(
+            `
+            select id, "transactionAddress", "walletIn", "walletOut", amount, "createdAt" from transaction as t
+            where t."walletOut" = $1
+            and t."walletIn" not in (select lower(address) from wallet)
```
jason-gerard, line 43: You can make these queries faster also by only getting the btc / eth wallets from the 2 sub queries

packages/btc-edge/src/services/wallets.service.ts
```diff
@@ -70,4 +71,25 @@ export class WalletsService {
...
+    async findOne(address: string, userId: number): Promise<Wallet> {
+        return this.walletRepository.findOne({ where: { address, userId } });
+    }
+
+    async delete(deleteWalletReq: DeleteWalletRequest): Promise<WalletWithBalance> {
+        const wallet = await this.findOne(deleteWalletReq.address, deleteWalletReq.id);
+        const [balance, count] = await Promise.all([
+            this.soChainGateway.getBalance(deleteWalletReq.address),
+            this.walletRepository.countBy({ address: deleteWalletReq.address }),
+            this.walletRepository.delete({ address: deleteWalletReq.address, userId: deleteWalletReq.id }),
```
jason-gerard, line 84: You will want to delete the wallet after getting the count, since we can't guarantee an ordering here, don't want to execute requests asynchronously that share resurces the same as with threads

packages/common/src/domain/entities/transaction.ts
```diff
@@ -21,6 +21,7 @@ export class Transaction {
     @CreateDateColumn({ type: "timestamp" })
     createdAt: Date;
 
+    @JoinTable()
```
jason-gerard, line 24: Was this not added before?

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -80,10 +80,10 @@ export class WalletsService {
         const [balance, count] = await Promise.all([
             this.alchemyNodeServiceFacade.getBalance(deleteWalletReq.address),
             this.walletRepository.countBy({ address: deleteWalletReq.address }),
+            this.walletRepository.delete({ address: deleteWalletReq.address, userId: deleteWalletReq.id }),
```
jason-gerard, line 83: Same reason as above mr incorrect optimizations

packages/btc-edge/src/services/transaction_watcher_service.ts
```diff
@@ -59,6 +59,15 @@ export class TransactionWatcherService {
             }),
         );
     }
+
+    async unsubscribeAddress(address: string): Promise<void> {
```
jason-gerard, line 63: Lovely

packages/common/src/domain/entities/transaction.ts
alexbara2000: It was added for transactionTags. Typeorm said it was only required for one table but by not having it here I am getting "Jointables" errors when I try to run my query from the transactionRepository.

packages/eth-edge/src/services/wallets.service.ts
alexbara2000: Did you mean mr Optimus Prime?

jason-gerard (approved):
Thank you optimus prime

PR #120, opened 2022-11-23
CRYP-175: delete tag added endpoint and gateway

Andre-Ibrahim:


packages/api/src/services/tags.service.ts
```diff
@@ -47,4 +48,17 @@ export class TagsService {
...
+        const userId = deleteTagRequest.id;
+        const tagName = deleteTagRequest.name;
+        const tag = await this.tagRepository.findOneBy({ userId, tagName });
+        if (!tag) {
+            throw new BadRequestException(ERROR_TAG_NOT_FOUND);
+        }
+
+        this.tagRepository.delete({ userId, tagName });
+
+        return tag;
```
jason-gerard, line 62: I think you want to do it like this, the other way is fine too just make sure you await the delete of the tag so the UI doesn't have any weird side effects

```suggestion
        return this.tagRepository.remove(tag);
```

jason-gerard (approved):
Nice and simple, just make sure you do one of the options I suggested. Its fine to run operations async only if the UI doesn't depend on it which for here it would

packages/api/src/services/tags.service.ts
```diff
@@ -47,4 +48,17 @@ export class TagsService {
...
+
+    async delete(deleteTagRequest: DeleteTagRequest): Promise<TransactionTag> {
+        const userId = deleteTagRequest.id;
+        const tagName = deleteTagRequest.name;
+        const tag = await this.tagRepository.findOneBy({ userId, tagName });
+        if (!tag) {
+            throw new BadRequestException(ERROR_TAG_NOT_FOUND);
+        }
+
+        this.tagRepository.delete({ userId, tagName });
```
alexbara2000, line 60: What happens if a transaction has a tag associated to it. Will this line delete the tag only from the transactionTag table or will it also delete the tag from the list of tags associated from a transaction. I have a feeling it is the former.

alexbara2000 (approved):

PR #125, opened 2022-11-28
CRYP-234: Add check to early return if wallet is already added

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-234

## Changes
- Add check to skip backfill if wallet already exists


Andre-Ibrahim (approved):
makes sense will reset my local database!

PR #127, opened 2022-11-30
CRYP-234: Update conditional check when skipping transactions

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-234


Andre-Ibrahim (approved):
thanks for the quick fix!

PR #128, opened 2022-11-30
CRYP-29: Do major minor pin instead of only major pin

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-29


alexbara2000 (approved):
Very nice

PR #130, opened 2022-12-02
CRYP-40: implement email notifications

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-40

## Changes
- Add notification abstract template method
- Add email notification implementation
- Add function call and integration to btc-edge, eth-edge will be done in another PR


packages/btc-edge/src/services/transaction_watcher_service.ts
```diff
@@ -43,6 +46,9 @@ export class TransactionWatcherService {
                 const txAddress = (res as WSTransaction).x.hash;
                 const transactions = await this.soChainGateway.getTransactionsByTxAddress(txAddress);
                 await this.transactionsRepository.save(transactions);
+
+                // Asynchronously send the notification to the users
+                this.notificationService.sendTransactionNotifications(transactions, getCurrencyType(txAddress));
```
Andre-Ibrahim, line 51: Why is this only in Btc-edge?

also just leaving a note that this needs to be only if user allows notification.

Andre-Ibrahim (approved):
from what I see there doesn't seem to be any setup for the mailer other than `yarn install` correct?

Everything else looks good just left a comment as a note for a future user story!

packages/common/src/utils/notifications/abstract_notification_service_template_method.ts
```diff
@@ -0,0 +1,59 @@
...
+                        // template method design pattern. This allows us to centralize all the pre-processing and
+                        // formatting in a single place while allowing for a module notification system. This can be
+                        // easily expanded to support sms and push notifications by simply extending this abstract
+                        // class and implementing only the specific functionality to send out a single notification
+                        // for that platform
+                        await this.sendNotification({
+                            to: wallet.user.email,
+                            title,
+                            body,
+                        });
```
alexbara2000, line 45: This code smells

alexbara2000 (approved):
Looks very good! The only thing, would be the nesting in the notification service but not a big deal.

packages/common/src/utils/notifications/abstract_notification_service_template_method.ts
jason-gerard: Yea the nesting could be improved upon thank you for your comment. I'll be updating it later on.

packages/btc-edge/src/services/transaction_watcher_service.ts
jason-gerard: Yep just adding a POC here, then in another pr ill add it in eth-edge and with the db updates etc

PR #133, opened 2022-12-05
CRYP-40: create endpoint to update notification prefs

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-40

## Changes
- Create new endpoint that allows updating a users attributes, this uses the active record design pattern through typeorm

## Dev notes
This will only be used right now to update a users notification preferences but can be used to update other attributes in the future.


Andre-Ibrahim (approved):

alexbara2000 (approved):
Looks good

PR #135, opened 2022-12-09
CRYP-41: added a screen for user to allow email notifications for transactions

Andre-Ibrahim:
IMG_1757
IMG_1756
IMG_1758


packages/client/screens/NotificationsScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+    async function toggleSwitch() {
+        usersGateway.update(
+            {
+                userId: user.id,
+                areNotificationsEnabled: !isEnabled,
+            },
+            token,
+        );
+
+        user.areNotificationsEnabled = !isEnabled;
```
jason-gerard, line 24: I don't think you can set the context like this, I wanna say you have to use the actual setter, did you try to back out to a diffent stack after changing the value to see if it persisted properly?

The user is from a set state so you cant mutate it.

packages/common/src/requests/update_user_request.ts
```diff
@@ -1,5 +1,8 @@
...
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface UpdateUserRequest extends InferType<typeof updateUserSchema> {}
+export type UpdateUserRequest = {
+    userId: number;
+    firstName?: string;
+    lastName?: string;
+    email?: string;
+    password?: string;
+    areNotificationsEnabled?: boolean;
+};
```
jason-gerard, line 8: Can you make this an interface still so that the auto class generator will pick up on it.
```suggestion
export interface UpdateUserRequest {
    userId: number;
    firstName?: string;
    lastName?: string;
    email?: string;
    password?: string;
    areNotificationsEnabled?: boolean;
}
```

packages/api/src/services/users.service.ts
```diff
@@ -47,7 +47,7 @@ export class UsersService {
         if (updateUserRequest.password) {
             user.password = updateUserRequest.password;
         }
-        if (updateUserRequest.areNotificationsEnabled) {
+        if (updateUserRequest.areNotificationsEnabled != null) {
```
jason-gerard, line 50: Thank you

jason-gerard (approved):

packages/client/screens/NotificationsScreen.tsx
Andre-Ibrahim: yup tested it the data persists, not sure how to use the setter for the authContext is there anywhere in the app where it's already done?

packages/client/screens/NotificationsScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+                <VStack style={{ maxWidth: 280, paddingRight: 75 }}>
+                    <Text>Trasactions</Text>
+                    <Text color="text.500" size={"footnote1"}>
+                        Send and receive transactions for all wallets
+                    </Text>
+                </VStack>
+                <Switch
+                    trackColor={{ false: "#767577", true: "#81b0ff" }}
+                    thumbColor={"#f4f3f4"}
+                    ios_backgroundColor="#3e3e3e"
```
domsec, line 41: Use colors present in the UI mockup.

```suggestion
                    trackColor={{ false: "#787880", true: "#0077E6" }}
                    thumbColor={"#FFFFFF"}
```

packages/client/screens/NotificationsScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+            </HStack>
+        </View>
+    );
+}
+
+const styles = StyleSheet.create({
+    view: {
+        flex: 1,
+        paddingHorizontal: 15,
+        paddingTop: 40,
```
domsec, line 54:
```suggestion
        paddingTop: 30,
```

packages/client/screens/NotificationsScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+            token,
+        );
+
+        user.areNotificationsEnabled = !isEnabled;
+        setIsEnabled(!isEnabled);
+    }
+
+    return (
+        <View style={styles.view}>
+            <HStack alignItems={"center"} space={19}>
```
domsec, line 30: For some reason native base styling doesn't treat space amount as pixels so we have to set it manually.

```suggestion
            <HStack alignItems={"center"} space={"15px"}>
```

packages/client/screens/NotificationsScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+                <FontAwesomeIcon icon={farArrowRightArrowLeft} size={26} />
```
domsec, line 31: This should be a light font weight and not regular

packages/client/components/icons/regular/farArrowRightArrowLeft.tsx
```diff
@@ -0,0 +1,18 @@
+import { IconDefinition, IconName, IconPrefix } from "@fortawesome/fontawesome-svg-core";
+
+export const farArrowRightArrowLeft: IconDefinition = {
```
domsec, line 3: Should be light font weight and not regular

packages/client/screens/NotificationsScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+                <VStack style={{ maxWidth: 280, paddingRight: 75 }}>
```
domsec, line 32: We shouldn't hard code the dimensions and padding right because it will cause problems on small and larger devices. Since you're using HStack as the parent with spacing then it'll be responsive by default.

domsec (approved):
So are we no longer doing push notifications and instead just email?

PR #137, opened 2023-01-01
CRYP-42: As a User, I want to set notification preferences, so that I only receive desired notifications

alexbara2000:
IMG_2357


jason-gerard (approved):
Looks good

Andre-Ibrahim (approved):

PR #140, opened 2023-01-09
CRYP-188: fix no transactions / wallets being displayed while loading

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-188

## Changes
Fixes an issue where the UI would display no wallets or no transactions to the user while the application was still loading the data from the API. Now the section will remain blank while the API request is still loading / in progress.


domsec (approved):
Looks good. 

We should probably in the future add a loading spinner instead of just keeping it blank to indicate to the user the transactions are being loaded and will be displayed shortly. What do you think @alexbara2000?

alexbara2000:
> Looks good.
> 
> We should probably in the future add a loading spinner instead of just keeping it blank to indicate to the user the transactions are being loaded and will be displayed shortly. What do you think @alexbara2000?

I like the idea especially if it takes a long time

alexbara2000 (approved):
Looks good

Andre-Ibrahim (approved):

PR #143, opened 2023-01-10
CRYP-41: Fix styling on notifications page

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-41

## Changes
- Fix typo
- Fix spacing on android
- Fix text spacing on all devices

## Dev note
- On Android the switch is smaller than the one on iOS but they take the same footprint so on Android the button is pushed to the left 10-15px more than on iOS, fixing this on Android would then make the iOS version look too far to the right. I've opted to just leave it as is for now, we can write some custom Android css rule to fix it if needed in the future.

## Screenshots
Screenshot_1673383616

### DIFF
--- a/packages/client/screens/NotificationsScreen.tsx
+++ b/packages/client/screens/NotificationsScreen.tsx
@@ -29,10 +29,10 @@ export default function NotificationsScreen() {
 
     return (
         <View style={styles.view}>
-            <HStack alignItems={"center"} space={"lg"}>
+            <HStack alignItems="center">
                 <FontAwesomeIcon icon={falArrowRightArrowLeft} size={26} />
-                <VStack>
-                    <Text>Trasactions</Text>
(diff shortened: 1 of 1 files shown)


domsec (approved):
Maybe make a note somewhere (bug ticket in Jira) so we don't forget and we can tackle it in a future sprint. I'm thinking the reason for this discrepancy is due to the toggle component styling in NativeBase.

narroarrow (approved):

jason-gerard:
> Maybe make a note somewhere (bug ticket in Jira) so we don't forget and we can tackle it in a future sprint. I'm thinking the reason for this discrepancy is due to the toggle component styling in NativeBase.

It uses the switch from React Native like it says too in the Figma mockups, I think the reason is that it doesn't use the NativeBase component and React Native doesn't handle the styling at all.

PR #148, opened 2023-01-12
CRYP-242: Refactor WalletDetailsScreen to simplify layout component structure

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-242

## Changes
- Refactor the layout components (VStack, HStack, and Box) used to simplify how the Wallet Details screen (WalletDetailsScreen.tsx) is built.
- Broke the original `RowItem` component into two separate list item components to handle both single and multiple line types used in various screens
- Refactored WalletQRCodeScreen to utilize newly created multiple list item component

## Dev Note
- With the addition of these new components we can simplify all other screens consisting of similar list items to use these components


jason-gerard (approved):
Cool nice refactor

Andre-Ibrahim (approved):
Thanks for the refactor!!

PR #153, opened 2023-01-13
CRYP-239: Fix the bottom navigation tab bar displaying on screens it should not

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-239

## Fix
- The bottom navigation tab bar displaying on screens it should not according to the UI mockups
- Also fixed an issue where dynamically removing the bottom navigation tab bar leaves behind a grey box on iOS devices

@jason-gerard Can you please test this on your emulator to ensure this bug is fixed as I'm only able to test this on an iOS device. Also would it be preferred to have the `useTabBarHook` function placed in a standalone file similar to `useCachedResources.tsx` in the `packages/client/hooks` directory?


jason-gerard (approved):
Yea if you can move it to a separate hooks file would be best. Ill test it out later today and leave another comment

jason-gerard:
@domsec just tested it out on android and all looks good to go

narroarrow (approved):

alexbara2000 (approved):
Separating your concerns is a very good idea, looks good!

domsec:
> @domsec just tested it out on android and all looks good to go

Perfect, i'll go ahead an create the separate hooks file tomorrow.

PR #156, opened 2023-01-14
CRYP-34: frontend for adding contacts

Andre-Ibrahim:
IMG_2039

### DIFF
--- a/packages/client/components/icons/light/falContactBook.tsx
+++ b/packages/client/components/icons/light/falContactBook.tsx
@@ -1,0 +1,18 @@
+import { IconDefinition, IconName, IconPrefix } from "@fortawesome/fontawesome-svg-core";
+
+export const falContactBook: IconDefinition = {
+    icon: [
+        // SVG viewbox width (in pixels)
+        512,
+        // SVG viewbox height (in pixels)
+        512,
+        // Aliases (not needed)
+        [],
+        // Unicode as hex value (not needed)
+        "f0f3",
+        // SVG path data
+        "M272 288h-64C163.8 288 128 323.8 128 368C128 376.8 135.2 384 144 384C152.8 384 160 376.8 160 368C160 341.5 181.5 320 208 320h64c26.51 0 48 21.49 48 48c0 8.836 7.164 16 16 16c8.837 0 16-7.164 16-16C352 323.8 316.2 288 272 288zM496 192H480v96h16C504.8 288 512 280.8 512 272v-64C512 199.2 504.8 192 496 192zM240 256c35.35 0 64-28.65 64-64s-28.65-64-64-64s-64 28.65-64 64S204.7 256 240 256zM240 160c17.64 0 32 14.36 32 32s-14.36 32-32 32s-32-14.36-32-32S222.4 160 240 160zM496 320H480v96h16c8.837 0 16-7.164 16-16v-64C512 327.2 504.8 320 496 320zM496 64H480v96h16C504.8 160 512 152.8 512 144v-64C512 71.16 504.8 64 496 64zM384 0H96C60.65 0 32 28.65 32 64v384c0 35.35 28.65 64 64 64h288c35.35 0 64-28.65 64-64V64C448 28.65 419.3 0 384 0zM416 448c0 17.64-14.36 32-32 32H96c-17.64 0-32-14.36-32-32V64c0-17.64 14.36-32 32-32h288c17.64 0 32 14.36 32 32V448z",
+    ],
+    iconName: "fa-contact-book" as IconName,
+    prefix: "fac-light" as IconPrefix,
+};
--- a/packages/client/components/icons/regular/farChevronUp.tsx
+++ b/packages/client/components/icons/regular/farChevronUp.tsx
@@ -1,0 +1,18 @@
+import { IconDefinition, IconName, IconPrefix } from "@fortawesome/fontawesome-svg-core";
+
+export const farChevronUp: IconDefinition = {
+    icon: [
+        // SVG viewbox width (in pixels)
+        512,
+        // SVG viewbox height (in pixels)
+        512,
+        // Aliases (not needed)
+        [],
+        // Unicode as hex value (not needed)
+        "f078",
+        // SVG path data
+        "M15.41 302.7l191.1-183.1C212 114.2 218 111.1 224 111.1s11.97 2.219 16.59 6.688l191.1 183.1c9.594 9.152 9.906 24.34 .7187 33.9c-9.125 9.625-24.38 9.938-33.91 .7187L224 169.2l-175.4 168c-9.5 9.219-24.78 8.906-33.91-.7187C5.502 327 5.814 311.8 15.41 302.7z",
+    ],
+    iconName: "far-chevron-up" as IconName,
+    prefix: "far-regular" as IconPrefix,
+};
--- a/packages/client/gateways/contacts_gateway.ts
+++ b/packages/client/gateways/contacts_gateway.ts
@@ -19,7 +19,7 @@ export class ContactsGateway extends AbstractApiGateway {
     }
 
     async createContact(req: CreateContactRequest, token: string): Promise<Contact> {
-        const path = `users/${req.userId}/contact`;
+        const path = `users/${req.userId}/contacts`;
         const headers = {
             Authorization: `Bearer ${token}`,
         };
--- a/packages/client/gateways/contacts_gateway.ts
+++ b/packages/client/gateways/contacts_gateway.ts
@@ -1,0 +1,29 @@
+import { GetContactsRequest } from "@cryptify/common/src/requests/get_contacts_request";
+import { AbstractApiGateway } from "./abstract_api_gateway";
+import { Method } from "@cryptify/common/src/utils/gateway/abstract_gateway";
+import { CreateContactRequest } from "@cryptify/common/src/requests/create_contact_request";
+import { Contact } from "@cryptify/common/src/domain/entities/contact";
+
+export class ContactsGateway extends AbstractApiGateway {
+    constructor() {
+        super();
+    }
+
+    async findAllContacts(req: GetContactsRequest, token: string): Promise<Contact[]> {
+        const path = `users/${req.id}/contacts`;
+        const headers = {
+            Authorization: `Bearer ${token}`,
+        };
+
+        return this.request<Contact[]>(Method.GET, headers, path, null);
+    }
+
+    async createContact(req: CreateContactRequest, token: string): Promise<Contact> {
+        const path = `users/${req.userId}/contact`;
+        const headers = {
+            Authorization: `Bearer ${token}`,
+        };
+
+        return this.request<Contact>(Method.POST, headers, path, req);
+    }
+}
--- a/packages/client/screens/AddContactScreen.tsx
+++ b/packages/client/screens/AddContactScreen.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { SettingsStackScreenProps } from "../types";
 import { View } from "../components/Themed";
-import { FieldArray, Formik, FormikErrors, FormikHelpers } from "formik";
+import { FieldArray, Formik, FormikErrors, FormikHelpers, FormikTouched } from "formik";
 import { Button, FormControl, HStack, Input, ScrollView, Text } from "native-base";
 import Collapsible from "react-native-collapsible";
 import { Pressable, StyleSheet } from "react-native";
@@ -20,7 +20,6 @@ import { ContactsGateway } from "../gateways/contacts_gateway";
 import { AuthContext } from "../components/contexts/AuthContext";
 
 export default function AddContactScreen(props: SettingsStackScreenProps<"ContactsSettingsScreen">) {
-
     const contactsGateway = new ContactsGateway();
 
     const { token, user } = React.useContext(AuthContext);
@@ -42,13 +41,16 @@ export default function AddContactScreen(props: SettingsStackScreenProps<"Contac
         handleChange: any;
         currencyType: CurrencyType;
         errors: FormikErrors<IValue>;
+        touched: FormikTouched<IValue>;
     };
 
-    function AddWalletFieldArray({ values, handleChange, currencyType, errors }: addWalletFieldArrayProps) {
+    function AddWalletFieldArray({ values, handleChange, currencyType, errors, touched }: addWalletFieldArrayProps) {
         const wallets = currencyType === CurrencyType.BITCOIN ? values.btcWallets : values.ethWallets;
         const walletListString = currencyType === CurrencyType.BITCOIN ? "btcWallets" : "ethWallets";
         const currencyIcon = currencyType === CurrencyType.BITCOIN ? faBitcoin : faEthereum;
         const iconColor = currencyType === CurrencyType.BITCOIN ? "#F7931A" : "#3C3C3D";
+        const currencyErrors = currencyType === CurrencyType.BITCOIN ? errors.btcWallets : errors.ethWallets;
+        const currencyTouched = currencyType === CurrencyType.BITCOIN ? touched.btcWallets : touched.ethWallets;
 
         const [isCollapsed, setIsCollapsed] = React.useState<boolean>(true);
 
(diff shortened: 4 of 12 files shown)


alexbara2000 (approved):
Looks very good

packages/client/gateways/contacts_gateway.ts
```diff
@@ -0,0 +1,29 @@
...
+        const path = `users/${req.id}/contacts`;
+        const headers = {
+            Authorization: `Bearer ${token}`,
+        };
+
+        return this.request<Contact[]>(Method.GET, headers, path, null);
+    }
+
+    async createContact(req: CreateContactRequest, token: string): Promise<Contact> {
+        const path = `users/${req.userId}/contact`;
```
jason-gerard, line 22: Should be plural in the controller as well

```suggestion
        const path = `users/${req.userId}/contacts`;
```

packages/client/screens/ContactsSettingsScreen.tsx
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+
+export default function ContactsSettingsScreen() {
```
jason-gerard, line 3: A better name, shouldn't be tied to the settings page

```suggestion
export default function ContactsListScreen() {
```

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+                } catch (e) {
+                    formikHelpers.setFieldError(
+                        `${walletListString}[${i}]`,
+                        ERROR_WALLET_ADDRESS_INVALID_FOR_CURRENCY(titleCase(currencyType)).split(":")[1],
+                    );
+                    hasError = true;
+                    return;
+                }
+            });
+        });
```
jason-gerard, line 172: This is very odd, you just need to loop over all the inputs and set the field error based on it, not sure what all the nesting is from. If you can't fix it atleast move it to a different function to hide it.

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+        try {
+            if (hasError) {
+                return;
+            }
```
jason-gerard, line 177: Everytime you set the hasError value to true you return right after so this doesn't need to be here

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+            const requestValues = { ...values };
+            requestValues.btcWallets = requestValues.btcWallets.filter((w) => w === "");
+            requestValues.ethWallets = requestValues.ethWallets.filter((w) => w === "");
```
jason-gerard, line 181: Why do you need to do this filter of empty values? The form shouldn't allow empty values to be added in the first pace

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+        btcWallets: [],
+    };
+
+    type addWalletFieldArrayProps = {
+        values: FormValuesType;
+        handleChange: any;
+        currencyType: CurrencyType;
+        errors: FormikErrors<FormValuesType>;
+        touched: FormikTouched<FormValuesType>;
+    };
```
jason-gerard, line 45: Would move these outside of the function, same file though

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+export default function AddContactScreen(props: SettingsStackScreenProps<"AddContactScreen">) {
+    const contactsGateway = new ContactsGateway();
+
+    const { token, user } = React.useContext(AuthContext);
+
+    type FormValuesType = {
+        contactName: string;
+        ethWallets: string[];
+        btcWallets: string[];
+    };
```
jason-gerard, line 31: Can't you use the type from the common?

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+        touched: FormikTouched<FormValuesType>;
+    };
+
+    function AddWalletFieldArray({ values, handleChange, currencyType, errors, touched }: addWalletFieldArrayProps) {
+        const wallets = currencyType === CurrencyType.BITCOIN ? values.btcWallets : values.ethWallets;
+        const walletListString = currencyType === CurrencyType.BITCOIN ? "btcWallets" : "ethWallets";
+        const currencyIcon = currencyType === CurrencyType.BITCOIN ? faBitcoin : faEthereum;
+        const iconColor = currencyType === CurrencyType.BITCOIN ? "#F7931A" : "#3C3C3D";
+        const currencyErrors = currencyType === CurrencyType.BITCOIN ? errors.btcWallets : errors.ethWallets;
+        const currencyTouched = currencyType === CurrencyType.BITCOIN ? touched.btcWallets : touched.ethWallets;
```
jason-gerard, line 53: There is a list I made called `currenciesDisplayData` which you can add too if needed but you can just say `currenciesDisplayData[0].icon` to get the bitcoin icon for example. Will remove all this code and bad use of ternary statements.

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+                        ) : (
+                            <FontAwesomeIcon style={styles.chevronIcon} icon={farChevronDown} size={18} />
+                        )}
+                    </HStack>
+                </Pressable>
+                {/* @ts-expect-error this is a known type error in the dependency a pr was made to address the issue but never merged */}
+                <Collapsible collapsed={isCollapsed}>
+                    <FieldArray
+                        name={walletListString}
+                        render={(arrayHelpers) => (
```
jason-gerard, line 80: Maybe what the list item actually is like wallets is a better name?

```suggestion
                        render={(wallets) => (
```

packages/client/screens/AddContactScreen.tsx
Andre-Ibrahim: Based on the specs we should just ignore empty fields

packages/client/screens/AddContactScreen.tsx
Andre-Ibrahim: The request requires a user id that why i need this type specifically for the form

packages/client/screens/AddContactScreen.tsx
Andre-Ibrahim: Need to do it for both eth and btc also there is two cases from that getCurrencyType i need to check one could return the wrong currency type the other would throw an error

packages/client/screens/AddContactScreen.tsx
Andre-Ibrahim: I thought thats how it should work too but it doesnt 😅 will take a second look at it maybe im missing something

jason-gerard: Its because above you are returning while in a map function so it returns from the inner closure

packages/client/screens/AddContactScreen.tsx
jason-gerard: Lol i don't think that what is meant

packages/client/screens/AddContactScreen.tsx
jason-gerard: Just add the user id as the default value instead of later on, you have it as soon as the page loads no need to dup this type just to add it in with the spread operator later

packages/client/screens/AddContactScreen.tsx
jason-gerard: The intialValues would stay if you add the userId, and then you can just delete the FormValuesType, and the addWalletFieldArrayProps probably doesnt need to exist either

packages/client/screens/AddContactScreen.tsx
jason-gerard: Basically you don't need any of this just use yup to validate the schema but if you just want to fix the current code then
- combine the 2 list together `[...values.btcWallets, ...values.ethWallets]` and foreach over that not a nested map, map is when there is a return value
- use the getCurrencyType inside the foreach to get your field string
- make a new function `isAddressValid` instead of catching the error from the current one
- remove the return statements
- remove hasError, there is a formik helper that will say if any of the error fields are active, if that true then just return early no need for this other value

Andre-Ibrahim: Not sure if im missing something but from this implementation we lose which input is which currencyType no?

Like if a user enters a eth address for a bitcoin wallet how do we tell the difference if there all merged into one list?

jason-gerard: You can use the getCurrencyType from the address to get the type at any point. you are already doing it

Andre-Ibrahim: Thats based on the value of the input the confusion happens if a user puts in a valid eth wallet address in a bitcoin field

jason-gerard: hmm yea you're right. idk why we even separate them. like at the point just add it anyway. but yea refactor if you like maybe just with a zipped for each because of that or don't refactor. it's all 1s and 0s in the end.

jason-gerard (approved):
Looks very good

packages/client/screens/AddContactScreen.tsx
domsec: Ignore all empty fields when passing to the backend (i.e., only validate and pass to backend non-empty fields). I could change the specs to make this clearer if it isn't.

Andre-Ibrahim: So my code makes sense right?

domsec: Isn't this code filtering to take all the empty fields when it should be the opposite?

packages/client/components/icons/light/falContactBook.tsx
```diff
@@ -0,0 +1,18 @@
+import { IconDefinition, IconName, IconPrefix } from "@fortawesome/fontawesome-svg-core";
+
+export const falContactBook: IconDefinition = {
```
domsec, line 3: Rename to `falAddressBook` to match the font awesome name

packages/client/components/icons/light/falContactBook.tsx
```diff
@@ -0,0 +1,18 @@
...
+        // SVG viewbox height (in pixels)
+        512,
+        // Aliases (not needed)
+        [],
+        // Unicode as hex value (not needed)
+        "f0f3",
+        // SVG path data
+        "M272 288h-64C163.8 288 128 323.8 128 368C128 376.8 135.2 384 144 384C152.8 384 160 376.8 160 368C160 341.5 181.5 320 208 320h64c26.51 0 48 21.49 48 48c0 8.836 7.164 16 16 16c8.837 0 16-7.164 16-16C352 323.8 316.2 288 272 288zM496 192H480v96h16C504.8 288 512 280.8 512 272v-64C512 199.2 504.8 192 496 192zM240 256c35.35 0 64-28.65 64-64s-28.65-64-64-64s-64 28.65-64 64S204.7 256 240 256zM240 160c17.64 0 32 14.36 32 32s-14.36 32-32 32s-32-14.36-32-32S222.4 160 240 160zM496 320H480v96h16c8.837 0 16-7.164 16-16v-64C512 327.2 504.8 320 496 320zM496 64H480v96h16C504.8 160 512 152.8 512 144v-64C512 71.16 504.8 64 496 64zM384 0H96C60.65 0 32 28.65 32 64v384c0 35.35 28.65 64 64 64h288c35.35 0 64-28.65 64-64V64C448 28.65 419.3 0 384 0zM416 448c0 17.64-14.36 32-32 32H96c-17.64 0-32-14.36-32-32V64c0-17.64 14.36-32 32-32h288c17.64 0 32 14.36 32 32V448z",
+    ],
+    iconName: "fa-contact-book" as IconName,
```
domsec, line 16:
```suggestion
    iconName: "fac-address-book" as IconName,
```

packages/client/components/icons/regular/farChevronUp.tsx
```diff
@@ -0,0 +1,18 @@
...
+        512,
+        // Aliases (not needed)
+        [],
+        // Unicode as hex value (not needed)
+        "f078",
+        // SVG path data
+        "M15.41 302.7l191.1-183.1C212 114.2 218 111.1 224 111.1s11.97 2.219 16.59 6.688l191.1 183.1c9.594 9.152 9.906 24.34 .7187 33.9c-9.125 9.625-24.38 9.938-33.91 .7187L224 169.2l-175.4 168c-9.5 9.219-24.78 8.906-33.91-.7187C5.502 327 5.814 311.8 15.41 302.7z",
+    ],
+    iconName: "far-chevron-up" as IconName,
+    prefix: "far-regular" as IconPrefix,
```
domsec, line 17:
```suggestion
    iconName: "fac-chevron-up" as IconName,
    prefix: "fac-regular" as IconPrefix,
```

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+                            <View>
+                                {wallets.map((wallet, i) => (
+                                    <View style={{ marginTop: 20 }} key={i}>
```
domsec, line 83: This is my fault for not showing what the UI should look like when there's multiple input fields. Make the marginTop 13 instead of 20. I'll update the UI mockup to represent this.

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+                                        </FormControl>
+                                    </View>
+                                ))}
+                                <View>
+                                    <Pressable
+                                        onPress={() => {
+                                            arrayHelpers.push("");
+                                        }}
+                                    >
+                                        <HStack style={{ marginTop: 14 }}>
```
domsec, line 119:
```suggestion
                                        <HStack style={{ marginTop: 13 }}>
```

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    return (
+        <View style={styles.view}>
+            <Formik initialValues={initialValues} onSubmit={onAddContactSubmit}>
+                {({ values, errors, touched, handleChange, submitForm }) => (
+                    <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
+                        <FormControl isInvalid={!!(errors.contactName && touched.contactName)}>
+                            <Input
```
domsec, line 198: The contact name form field should be focused on page load. You can copy the code needed from `AddTagsScreen`. If you have any issues let me know and I'll help you out 😄  .

domsec (approved):

packages/client/screens/AddContactScreen.tsx
Andre-Ibrahim: No it filters out all the empty strings that where allowed to go through by the form right before the api request

domsec: Then it's fine. Clearly i'm tired :sweat_smile:

domsec: > Why do you need to do this filter of empty values? The form shouldn't allow empty values to be added in the first pace

Reading this over a few times @jason-gerard has a point. Is there a way in Formik to ignore empty fields on form submission that way we avoid doing this filtering before the api request?

Andre-Ibrahim: Dont think so i think he thought it would be an error if it was empty no point to mess with formik like that to save a couple lines of code will probably make it more convoluted

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+                                value={values.contactName}
+                                onChangeText={handleChange("contactName")}
+                                placeholder="Name"
+                                maxLength={20}
+                                keyboardType={"ascii-capable"}
+                                testID="contactNameInput"
+                            />
+                            <FormControl.ErrorMessage>{errors.contactName}</FormControl.ErrorMessage>
+                        </FormControl>
+                        <AddWalletFieldArray
```
domsec, line 208: Looks like placeholder value is missing.

packages/client/screens/AddContactScreen.tsx
```diff
@@ -0,0 +1,258 @@
...
+                            values={values}
+                            handleChange={handleChange}
+                            currencyType={CurrencyType.BITCOIN}
+                            errors={errors}
+                            touched={touched}
+                        />
+                        <AddWalletFieldArray
```
domsec, line 215: Looks like placeholder value is missing

domsec (approved):

packages/client/screens/AddContactScreen.tsx
domsec: Then its fine with me.

PR #158, opened 2023-01-15
CRYP-232: transaction history report gen frontend

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-232

## Screenshots
Screenshot_1673813710
Screenshot_1673813712
Screenshot_1673814195
Screenshot_1673814200


packages/client/navigation/index.tsx
```diff
@@ -244,6 +246,54 @@ function HomeStackScreen({ navigation, route }: { route: RouteProp<any, any>; na
...
+                    // TODO refactor to reduce code duplication and unify with other screens
+                    headerLeft: () => (
+                        <Pressable
+                            onPress={() => {
+                                navigation.goBack();
+                            }}
+                        >
+                            <FontAwesomeIcon icon={farArrowLeft} color="#404040" size={22} />
+                        </Pressable>
+                    ),
```
domsec, line 270: You can remove this because it's fixed in https://github.com/SOEN-490-Capstone/Cryptify/pull/159

packages/client/navigation/index.tsx
```diff
@@ -244,6 +246,54 @@ function HomeStackScreen({ navigation, route }: { route: RouteProp<any, any>; na
...
```
domsec, line 294: Remove this as well

packages/client/screens/reports/ReportSelectionScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export default function ReportSelectionScreen({ route, navigation }: HomeStackScreenProps<"ReportSelectionScreen">) {
+    return (
+        <View style={styles.view}>
+            <VStack space="15px">
+                <Pressable
+                    onPress={() =>
+                        navigation.navigate("TransactionHistoryReportScreen", {
+                            walletAddress: route.params.walletAddress,
+                            walletName: route.params.walletName,
```
domsec, line 18: Would it be better to pass the wallet object instead?

packages/client/screens/reports/ReportSelectionScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+            </VStack>
+        </View>
+    );
+}
+
+const styles = StyleSheet.create({
+    view: {
+        flex: 1,
+        paddingHorizontal: 15,
+        paddingTop: 10,
```
domsec, line 42:
```suggestion
        paddingTop: 20,
        paddingBottom: 15,
```

packages/client/screens/reports/TransactionHistoryReportScreen.tsx
```diff
@@ -0,0 +1,213 @@
...
+import { FileType } from "@cryptify/common/src/validations/create_transaction_history_report_schema";
+
+type RadioProps = {
+    value: string;
+    setValue: React.Dispatch<React.SetStateAction<string>>;
+    options: string[];
+    isDisabled?: boolean;
+};
+
+export default function TransactionHistoryReportScreen({
```
domsec, line 21: Since the transaction history form is practically identical to the transaction list filter screen from a ui component perspective, would it be possible to create components for `filter by transaction` and `filter by date` to avoid code duplication? This way the custom data logic for instance is similar for both screens (they are already the same specs)

domsec (approved):
Looks good, see comment about possibly refactoring the filter screen and whether its feasible to do that this sprint or push it to a future one.

packages/client/gateways/reports_gateway.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class ReportsGateway extends AbstractApiGateway {
+    constructor() {
+        super();
+    }
+
+    async createTransactionHistoryReport(req: CreateTransactionHistoryReportRequest, token: string): Promise<void> {
+        const path = `users/${req.userId}/reports/transaction-history`;
+        const headers = {
+            Authorization: `Bearer ${token}`,
```
narroarrow, line 13: What does this do / What is the purpose of adding 'Bearer'?

narroarrow (approved):

packages/client/gateways/reports_gateway.ts
jason-gerard: This is for the JWT authentication

packages/client/navigation/index.tsx
jason-gerard: Can you merge that in, then i'll fix it here

packages/client/screens/reports/ReportSelectionScreen.tsx
jason-gerard: It would be better but since it wasn't written like that to start with I dont have access to the entire wallet and I don't want to refactor all thoe components in case I break something

packages/client/navigation/index.tsx
domsec: Merge is done

packages/client/screens/reports/ReportSelectionScreen.tsx
domsec: Makes sense, I'll take care of it when working on displaying the wallet name in the transaction details screen.

packages/client/screens/reports/TransactionHistoryReportScreen.tsx
jason-gerard: It would be good but the actual implementation of the 2 screens is different since one makes an API call and the other updates the transactions list so its just a but more work than i'd like to put in now. I think Andre wanted to refactor / completely change the implementation of the filters anyways so we can probably do this at the same time.

domsec: Perfect. Lets add a note to the associated ticket so he's aware of this refactor as well.

PR #168, opened 2023-01-26
CRYP-232: add performance test for creating txn history report

jason-gerard:
https://cryptify.atlassian.net/browse/CRYP-232


narroarrow (approved):
Nice

Andre-Ibrahim (approved):

alexbara2000 (commented):
Very good tasking and I like k6 in general because of its simplicity

PR #183, opened 2023-02-01
CRYP-300: refactor transaction tag to tag

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-177
JIRA: https://cryptify.atlassian.net/browse/CRYP-300


domsec (approved):
Good idea renaming transaction tags since now it'll be more clear. I think it's probably best to create a new Jira ticket for this so we can easily keep track of refactors for the release 3 report.

jason-gerard:
> Good idea renaming transaction tags since now it'll be more clear. I think it's probably best to create a new Jira ticket for this so we can easily keep track of refactors for the release 3 report.

Ok ill do that

narroarrow (approved):
Nice

PR #184, opened 2023-02-01
CRYP-247: update db schema for pro mode

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-247


narroarrow (approved):

Andre-Ibrahim (approved):
Looks good but shouldn't the create also set that field or do we plan on making a user make a  basic account first then upgrade if they want?

jason-gerard:
> Looks good but shouldn't the create also set that field or do we plan on making a user make a basic account first then upgrade if they want?

Based on the flow there isn't an option right now to set the mode upon create so all accounts are just defaulted to Basic. Through the settings users can update their account to a Pro mode using the update API. Basically working the same way as notification preferences are.

PR #187, opened 2023-02-02
CRYP-204: delete user account

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-204

## Changes
- Add new API that deletes user and all related resources in parallel but without any race conditions
- Implement delete account button on frontend
- Implement re-routing of user to welcome screen after account has been deleted

## Screenshots
Screenshot_1675311114
Screenshot_1675311115


narroarrow (approved):

Andre-Ibrahim (approved):
looks great!

packages/api/src/services/users.service.ts
```diff
@@ -56,4 +65,31 @@ export class UsersService {
...
+
+        const [wallets] = await Promise.all([
+            this.walletsService.findAll({ id }),
+            this.tagRepository.delete({ userId: id }),
+            this.contactRepository.delete({ userId: id }),
+        ]);
+
+        // Manually delete the wallets through the edge services so that any cleanup process that needs to happen with
+        // the transactions doesn't need to be replicated here
+        await Promise.all(wallets.map(({ address }) => this.walletsService.delete({ id, address })));
```
alexbara2000, line 85: what if we don't await and let it run in the background?

alexbara2000 (approved):
Looks good!

packages/api/src/services/users.service.ts
jason-gerard: Because then youd get a db foreign key error when you try to delete a user when the wallet still references them. There is also no reason to do it in the background anyways, if the frontend doesn't want to wait it doesnt have too the API shouldn't decide that

PR #188, opened 2023-02-02
CRYP-191: Edit a contact

alexbara2000:
Jira: https://cryptify.atlassian.net/browse/CRYP-191 

Added backend for editing a contact
Changed the AddContact screen to allow for editing
The screen therefore looks the same, but has the extra functionalities.


packages/client/screens/contacts/ContactsListScreen.tsx
```diff
@@ -50,7 +50,11 @@ export default function ContactsListScreen({ route, navigation }: Props) {
 
     async function onSubmit(contactName: string) {
         if (!route.params.prefilledWalletAddress) {
-            return;
+            const contacts = await contactsGateway.findSpecificContacts({ id: user.id, name: contactName }, token);
+            navigation.navigate("AddContactScreen", {
+                prefilledWalletAddress: undefined,
+                contacts: contacts,
+            });
```
jason-gerard, line 57: Instead of doing this  just grab the list of "contacts" from the api call already made based on the contact name. This avoids the extra API call. Then when the add wallet form is submited if its in edit mode filter out the list of wallet address with the ones passed in as props so you can remove that weird deletion of all contacts in the API.

jason-gerard (approved):

packages/client/screens/contacts/ContactsListScreen.tsx
alexbara2000: @jason-gerard take a look now. Let me know of any naming changes you want too.

packages/api/src/services/contacts.service.ts
```diff
@@ -15,19 +16,38 @@ export class ContactsService {
         return this.contactRepository.find({ where: { userId }, order: { contactName: "ASC" } });
     }
 
+    async findSpecific(userId: number, contactName: string) {
+        return this.contactRepository.find({ where: { userId, contactName } });
+    }
+
+    async delete(userId: number, contactName: string) {
+        return this.contactRepository.delete({ userId: userId, contactName: contactName });
+    }
```
jason-gerard, line 25: Dont need this anymore

packages/api/src/controllers/contacts.controller.ts
```diff
@@ -21,6 +23,13 @@ export class ContactsController {
         return this.contactsService.findAll(getContactsRequest.id);
     }
 
+    @UseGuards(JwtAuthGuard, CanAccessResourceGuard)
+    @Get("users/:id/contacts/:name")
+    async findSpecific(@Param() params: GetSpecificContactsRequest): Promise<Contact[]> {
```
jason-gerard, line 28: Standard naming convention would just be this

```suggestion
    async find(@Param() params: GetSpecificContactsRequest): Promise<Contact[]> {
```

packages/api/src/services/contacts.service.ts
```diff
@@ -15,19 +16,38 @@ export class ContactsService {
         return this.contactRepository.find({ where: { userId }, order: { contactName: "ASC" } });
     }
 
+    async findSpecific(userId: number, contactName: string) {
```
jason-gerard, line 19: Name and return type

```suggestion
    async find(userId: number, contactName: string): Promise<Contact> {
```

packages/api/src/services/contacts.service.ts
```diff
@@ -15,19 +16,38 @@ export class ContactsService {
...
                 userId,
                 contactName,
                 walletAddress: addr,
             }),
         );
 
-        await this.contactRepository.insert(contacts);
-        return contacts;
+        await this.contactRepository.insert(contactsToAdd);
+        return contactsToAdd;
```
jason-gerard, line 51: Just to make it uniform

```suggestion
        return [...contactsToAdd, ...contactsToDelete];
```

packages/client/gateways/contacts_gateway.ts
```diff
@@ -18,6 +19,15 @@ export class ContactsGateway extends AbstractApiGateway {
         return this.request<Contact[]>(Method.GET, headers, path, null);
     }
 
+    async findSpecificContacts(req: GetSpecificContactsRequest, token: string): Promise<Contact[]> {
```
jason-gerard, line 22:
```suggestion
    async findContacts(req: GetSpecificContactsRequest, token: string): Promise<Contact[]> {
```

packages/api/src/controllers/contacts.controller.ts
```diff
@@ -21,6 +23,13 @@ export class ContactsController {
         return this.contactsService.findAll(getContactsRequest.id);
     }
 
+    @UseGuards(JwtAuthGuard, CanAccessResourceGuard)
+    @Get("users/:id/contacts/:name")
```
jason-gerard, line 27:
```suggestion
    @Get("/users/:id/contacts/:name")
```

packages/common/src/validations/get_specific_contacts_schema.ts
```diff
@@ -0,0 +1,6 @@
+import * as yup from "yup";
+
+export const getSpecificContactsSchema = yup.object({
```
jason-gerard, line 3: Update file name too

```suggestion
export const getContactSchema = yup.object({
```

packages/common/src/requests/get_specific_contacts_request.ts
```diff
@@ -0,0 +1,5 @@
+import { InferType } from "yup";
+import { getSpecificContactsSchema } from "../validations/get_specific_contacts_schema";
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface GetSpecificContactsRequest extends InferType<typeof getSpecificContactsSchema> {}
```
jason-gerard, line 5: File name too

```suggestion
export interface GetContactRequest extends InferType<typeof getSpecificContactsSchema> {}
```

packages/client/screens/contacts/ContactsListScreen.tsx
jason-gerard: Looks much better, could still completely remove this API call + endpoint by just searching through the entire contacts list that is already fetched on this page. But if you update the naming of it like I suggested then it'll be fine enough.

jason-gerard (approved):

packages/api/src/services/contacts.service.ts
alexbara2000: I was going to leave it there for the person in charge of deleting contacts but I will let them figure it out :)

jason-gerard: oh yea we could leave it then if you want

packages/common/src/validations/get_specific_contacts_schema.ts
alexbara2000: There is already a file called that for when we get all contacts.

jason-gerard: `getContactSchema` != `getContactsSchema`

Andre-Ibrahim (approved):
you need to update the unit test since you updated the schema, other than that looks good!

nothing crazy probably a type error so you can just make those fields empty arrays. 
Also i believe that even if you put notRequired in yup when we convert it into a typescript type its a required field. Workaround would be to write out the type

PR #189, opened 2023-02-02
CRYP-206: Add cahcing to `findAll` wallets API

jason-gerard:
JIRA: https://cryptify.atlassian.net/browse/CRYP-206

## Changes
- Add a cache with a TTL of 10 second based on req route

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -20,6 +20,7 @@
         "@nestjs/platform-express": "^9.0.0",
         "@nestjs/typeorm": "^9.0.1",
         "bcrypt": "^5.0.1",
+        "cache-manager": "^5.1.4",
         "passport": "^0.6.0",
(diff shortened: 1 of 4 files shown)


narroarrow (approved):

alexbara2000 (approved):
Looks good, are we planning on caching everything?

jason-gerard:
> Looks good, are we planning on caching everything?

Yea it was simple enough to apply it to everything. Each endpoint can have the cache TTL set independently as well, so if we are seeing stale data it can be lowered as needed. 10 sec is a pretty reasonable standard as for the most part the data isn't changing all that often. Only GET requests are cached of course as well, the way its setup all new GET endpoints will be automatically cached without needing to do anything else

Andre-Ibrahim (approved):
I like it!

PR #190, opened 2023-02-03
CRYP-227: added gasLimit and gasPrice to Transaction entity

Andre-Ibrahim:


packages/eth-edge/src/services/alchemy_node_facade.service.ts
```diff
@@ -49,16 +49,23 @@ export class AlchemyNodeServiceFacade {
...
+                const transaction = await this.alchemy.getTransaction(transfer.hash);
+                return {
+                    id: -1,
+                    transactionAddress: transfer.hash,
+                    walletIn: transfer.to,
+                    contactIn: null,
+                    walletOut: transfer.from,
+                    contactOut: null,
+                    amount: Web3.utils.toWei(normalizeCurrency(transfer.value), "ether"),
+                    gasPrice: transaction.gasPrice.toString(),
```
jason-gerard, line 63: Are these values in wei already? because we will want to store them as wei and convert to eth later like we do for the transaction value

jason-gerard (approved):

narroarrow (approved):

packages/eth-edge/src/services/alchemy_node_facade.service.ts
Andre-Ibrahim: gasPrice is, gasLimit is not in ETHEREUM it is a the limit of unit gas that a transaction can take for example it would be something like "21 000 Units"

PR #192, opened 2023-02-04
CRYP-228: added ethereum transaction fee amount

Andre-Ibrahim:
image

### DIFF
--- a/packages/client/components/transaction-details/TransactionDetails.tsx
+++ b/packages/client/components/transaction-details/TransactionDetails.tsx
@@ -41,7 +41,8 @@ export function TransactionDetails({ txn, wallet, navigation, otherDetails }: Pr
     const [transaction, setTransaction] = React.useState<Transaction>(txn);
     const [transactionTags, setTransactionTags] = React.useState<Tag[]>([]);
     const isIncomingTransaction = wallet.address == transaction.walletIn;
-    const tranasctionFee = wallet.currencyType === CurrencyType.ETHEREUM? (+transaction.gasLimit)* +transaction.gasPrice : 0;
+    const tranasctionFee =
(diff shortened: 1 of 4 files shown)


packages/client/components/transaction-details/TransactionDetails.tsx
```diff
@@ -258,7 +261,9 @@ export function TransactionDetails({ txn, wallet, navigation, otherDetails }: Pr
             />
             <MultiLineListItem
                 label={"Fee"}
-                value={`${getFormattedAmount(String(0), wallet.currencyType)} ${typeToISOCode[wallet.currencyType]}`}
+                value={`${getFormattedAmount(String(tranasctionFee), wallet.currencyType)} ${
```
jason-gerard, line 264: I know its not your code but those type casting functions are no good

```suggestion
                value={`${getFormattedAmount(tranasctionFee.toString(), wallet.currencyType)} ${
```

jason-gerard (approved):
Why all the .toLowerCase()? does our APIs not respect casing?

Andre-Ibrahim:
> Why all the .toLowerCase()? does our APIs not respect casing?

I believe alchemy sends all the addresses in lowercase so that list was empty since the filter was never set to true

narroarrow (approved):

domsec (approved):
Why is the `Other Details` section no longer showing in the transaction details screen?

jason-gerard:
> Why is the `Other Details` section no longer showing in the transaction details screen?

Account probably isn't pro mode

domsec:
> > Why is the `Other Details` section no longer showing in the transaction details screen?
> 
> Account probably isn't pro mode

I just realized that is probably the case lol.

PR #194, opened 2023-02-13
CRYP-192: added backend and gateway to delete a contact

Andre-Ibrahim:


packages/api/src/services/contacts.service.ts
```diff
@@ -45,4 +46,14 @@ export class ContactsService {
         await this.contactRepository.insert(contactsToAdd);
         return [...contactsToAdd, ...contactsToDelete];
     }
+
+    async delete(deleteContactRequest: DeleteContactRequest): Promise<Contact[]> {
+        const { id, name } = deleteContactRequest;
+        const contacts = await this.contactRepository.findBy({ userId: id, contactName: name });
+        if (contacts) {
```
jason-gerard, line 53: Do you even need this if statement?

jason-gerard (approved):

packages/api/src/services/contacts.service.ts
Andre-Ibrahim: Avoiding an internal server error that would be thrown if we try to delete a contact that doesnt exist

narroarrow (approved):

PR #196, opened 2023-02-17
CRYP-192: Added delete contact button

PolaFarid:
IMG_2746
IMG_2747


packages/client/components/contacts/ContactsForm.tsx
```diff
@@ -199,4 +227,8 @@ const styles = StyleSheet.create({
     addContactButton: {
         marginTop: "auto",
     },
+    deleteContactButton: {
+        marginTop: "auto",
+        color: "white",
+    },
```
jason-gerard, line 233: I don't think you need to set the color here, it should do anything and the margin top doesn't follow the mockups. There should be some fixed pixel margin applied. This will just push the button down

packages/client/components/contacts/ContactsForm.tsx
```diff
@@ -181,6 +196,19 @@ export default function ContactsForm(props: Props) {
...
+                            _text={{ color: "error.500" }}
+                            onPress={handleDeleteContact}
+                            testID="deleteContactButton"
+                            style={styles.deleteContactButton}
+                        >
+                            Delete contact
+                        </Button>
+                    ) : (
+                        <></>
+                    )}
```
jason-gerard, line 211: You can also switch it from a ternary because there is no negative case
```suggestion
                    {props.contact && (
                        <Button
                            variant="outline"
                            _text={{ color: "error.500" }}
                            onPress={handleDeleteContact}
                            testID="deleteContactButton"
                            style={styles.deleteContactButton}
                        >
                            Delete contact
                        </Button>
                    )}
```

packages/client/components/contacts/ContactsForm.tsx
```diff
@@ -131,6 +131,21 @@ export default function ContactsForm(props: Props) {
...
+    function handleDeleteContact(): void {
+        Alert.alert("Do you want to delete this contact?", "You cannot undo this action.", [
+            { text: "Cancel", style: "cancel" },
+            {
+                text: "Delete",
+                style: "destructive",
+                onPress: async () => {
+                    await contactsGateway.deleteContact({ id: user.id, name: props.contact?.contactName }, token);
+                    props.navigation.goBack();
+                    props.navigation.goBack();
```
jason-gerard, line 143: I'd wrap this in an if statement so there is no chance of sending a null value.

```suggestion
                    if (props.contact) {
                        await contactsGateway.deleteContact({ id: user.id, name: props.contact.contactName }, token);
                        props.navigation.goBack();
                        props.navigation.goBack();
                    }
```

packages/client/components/contacts/ContactsForm.tsx
domsec: No need to set color since it is automatically set by the Button theme for the outline variant button (which you set correctly).

packages/client/components/contacts/ContactsForm.tsx
domsec: Add `marginTop={"35px"}` to fix the spacing and remove `styles.deleteContactButton`.

jason-gerard (approved):
LGTM

Andre-Ibrahim (approved):
nice!

PolaFarid:
Waiting for @domsec review

PR #206, opened 2023-03-09
CRYP-203: edit account information, name, email, password

Andre-Ibrahim:


packages/api/src/services/authentication.service.ts
```diff
@@ -33,6 +37,16 @@ export class AuthenticationService {
         return this.signToken(user);
     }
 
+    async isMyPassword(password: string, userId: number): Promise<string> {
```
jason-gerard:
```suggestion
    async verify(password: string, userId: number): Promise<string> {
```

packages/api/src/services/authentication.service.ts
```diff
@@ -33,6 +37,16 @@ export class AuthenticationService {
...
+        const user = await this.usersService.findOneById(userId);
+
+        return await bcrypt.compare(password, user.password);
+    }
+
+    async createNewPassword(password: string): Promise<string> {
```
jason-gerard, line 46:
```suggestion
    async encode(password: string): Promise<string> {
```

packages/client/components/list/MultiLineListItem.tsx
```diff
@@ -29,7 +29,7 @@ export default function MultiLineListItem({
             </Text>
             {altValue && <Text flex={1}>{altValue}</Text>}
             <HStack space="10px">
-                <Text flex={1}>{value}</Text>
```
jason-gerard, line 32: Why did you change this? Effects a lot of other places

packages/client/screens/account/AccountEmailScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+                formikHelpers.setFieldError("confirmEmail", error.message);
+            }
+        }
+    }
+
+    return (
+        <View style={styles.view}>
+            <Formik initialValues={intitialValues} validationSchema={updateUserSchema} onSubmit={handleUpdate}>
+                {({ values, errors, touched, handleChange, submitForm }) => (
+                    <VStack space={4} marginTop={5}>
```
jason-gerard, line 56: These dont reference pixel values btw you need to do `space={"4px"}` if that was what you wanted

packages/client/screens/account/AccountNameScreen.tsx
```diff
@@ -0,0 +1,115 @@
...
+                                disabled={
+                                    intitialValues.firstName === values.firstName &&
+                                    intitialValues.lastName === values.lastName
+                                }
+                                style={
+                                    intitialValues.firstName === values.firstName &&
+                                    intitialValues.lastName === values.lastName
+                                        ? styles.ButtonDisabled
+                                        : styles.Button
+                                }
```
jason-gerard, line 81: You aren't using the right props, this is how it should be. The disabled styles are already defined

```suggestion
                                isDisabled={
                                    intitialValues.firstName === values.firstName &&
                                    intitialValues.lastName === values.lastName
                                }
                                style={styles.Button}
```

packages/common/src/validations/update_user_schema.ts
```diff
@@ -5,18 +5,19 @@ export const updateUserSchema = yup.object({
...
         .string()
-        .matches(new RegExp("^[A-Za-z]+"), "First name must only contain alphabetic characters.")
+        .matches(new RegExp("^[A-Za-z]+$"), "First name must only contain alphabetic characters.")
         .min(0)
         .max(32, "First name must be 32 characters or less.")
         .nullable(),
     lastName: yup
         .string()
-        .matches(new RegExp("^[A-Za-z]+"), "Last name must only contain alphabetic characters.")
+        .matches(new RegExp("^[A-Za-z]+$"), "Last name must only contain alphabetic characters.")
```
jason-gerard, line 14: What is this?

packages/api/src/services/users.service.ts
```diff
@@ -53,12 +56,18 @@ export class UsersService {
...
+                throw new ForbiddenException(ERROR_CURRENT_PASSWORD_INCORRECT);
+            }
+            user.password = await this.authService.createNewPassword(updateUserRequest.newPassword);
         }
         if (updateUserRequest.areNotificationsEnabled != null) {
             user.areNotificationsEnabled = updateUserRequest.areNotificationsEnabled;
         }
+        if (updateUserRequest.email != null) {
+            user.email = updateUserRequest.email;
+        }
```
jason-gerard, line 70: Confirm email?

packages/common/src/validations/update_user_schema.ts
Andre-Ibrahim: it will still match the first part of the string like if you had Jason! it would match Jason, this makes sure that it validates the entire word

packages/api/src/services/users.service.ts
Andre-Ibrahim: Confirm email is handle by the schema validation same as confirm password

packages/common/src/validations/update_user_schema.ts
jason-gerard: Why do we want that?

packages/client/components/list/MultiLineListItem.tsx
Andre-Ibrahim: the text gets wrapped in weird way over 3 characters with it

packages/api/src/services/users.service.ts
jason-gerard: That's pretty cool boy

packages/client/components/list/MultiLineListItem.tsx
jason-gerard: But is it going to mess up the styling in all the other places it is used?

packages/common/src/validations/update_user_schema.ts
Andre-Ibrahim: without it it would still pass names with special characters because it matches something

packages/client/components/list/MultiLineListItem.tsx
Andre-Ibrahim: will take a look before merging the other pages

PolaFarid (approved):

packages/client/components/list/MultiLineListItem.tsx
Andre-Ibrahim: you where right i put a minWidth on it so it doesn't look weird instead and doubled checked all pages that use it which looked fine

PR #209, opened 2023-03-09
CRYP-199: As a User, I want to filter my Bitcoin wallet transactions by contact, so that I can search for specific transaction senders and receivers

alexbara2000:


narroarrow (approved):

alexbara2000:
image

image

image

alexbara2000:
image

packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -44,14 +52,20 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
                     setFilterByDate(filtersByDate[0]);
                     route.params.setFilters([filtersByTransaction[0], filtersByDate[0]]);
                     route.params.setIsUsingSavedFilter(false);
+                    route.params.setContactFilters([]);
                     setIsFilterSaved(false);
+                    navigation.goBack();
```
jason-gerard, line 57: I'm not sure the reset link should make the nav go back to the txn list, can you double check this?

packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -132,9 +146,41 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
...
+            <Box marginTop="20px" />
+            {!(
+                filterByTransaction === filtersByTransaction[0] &&
+                filterByDate === filtersByDate[0] &&
+                filterByContact.length === 0
+            ) &&
                 (!isFilterSaved ||
-                    !(filterByTransaction === route.params.filters[0] && filterByDate === route.params.filters[1])) && (
+                    (!(filterByTransaction === route.params.filters[0] && filterByDate === route.params.filters[1]) &&
+                        filterByContact.length === 0)) && (
```
jason-gerard, line 183: I think it should be like this no? where its inside the negation. I might be wrong though if you tested all the different cases, the logic gets confusing
```suggestion
                    (!(filterByTransaction === route.params.filters[0] && filterByDate === route.params.filters[1] && filterByContact.length === 0))) && (
```

packages/client/services/filter_service.ts
```diff
@@ -70,5 +71,12 @@ export function filterTransctions(
         }
     });
 
+    if (contactFilters.length !== 0) {
+        transactions = transactions.filter(
+            (transaction) =>
+                contactFilters.includes(transaction.contactIn?.contactName ?? "") ||
+                contactFilters.includes(transaction.contactOut?.contactName ?? ""),
```
jason-gerard, line 78: The logic right now will include transactions that aren't associated with a contact which isn't right. If contact name is null then `contactFilters.includes("")` will be true always. Just check before the includes if one is present, if not then shortcircuit

```suggestion
                transaction.contactIn?.contactName && contactFilters.includes(transaction.contactIn.contactName) ||
                transaction.contactOut?.contactName && contactFilters.includes(transaction.contactOut.contactName),
```

jason-gerard (approved):
Pretty good work with the bad initial code base, a lot of technical debt here. Shouldn't be to hard to make it work with ETH probably just some checks to make sure everything works and then its a very similar thing to get it to work with tags now.

Andre-Ibrahim (approved):
Looks good

packages/client/services/filter_service.ts
alexbara2000: that was kinda the idea. I wanted to include transactions without a contact. But looking back it does not make sense

PR #211, opened 2023-03-11
CRYP-37: As a User, I want to filter my Ethereum wallet transactions by tag, so that I can search for specific transaction characteristics

alexbara2000:
image

image

image

### DIFF
--- a/packages/client/screens/filters/FilterScreen.tsx
+++ b/packages/client/screens/filters/FilterScreen.tsx
@@ -11,6 +11,7 @@ import SaveFilterActionSheet from "../../components/SaveFilterActionSheet";
 import { fasBookmark } from "../../components/icons/solid/fasBookmark";
 import { farChevronRight } from "../../components/icons/regular/farChevronRight";
 import { useIsFocused } from "@react-navigation/native";
+import TagItem from "../../components/tags/TagItem";
 
 export default function FilterScreen({ route, navigation }: HomeStackScreenProps<"FilterScreen">) {
     const isFocused = useIsFocused();
@@ -25,6 +26,7 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
     const [filterByDate, setFilterByDate] = React.useState(route.params.filters[1] || filtersByDate[0]);
 
     const [filterByContact, setFilterByContact] = React.useState<string[]>([]);
+    const [filterByTag, setFilterByTag] = React.useState<string[]>([]);
 
     type RadioProps = {
         value: string;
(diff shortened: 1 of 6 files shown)


packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -180,20 +188,82 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
...
+                            <HStack space={"10px"} alignItems={"center"}>
+                                <Text size={"subheadline"} fontWeight={"semibold"}>
+                                    {tag}
+                                </Text>
+                            </HStack>
+                        </Pressable>
+                    ))}
+                    <FontAwesomeIcon icon={farChevronRight} style={styles.chevronRightIcon} size={16} />
+                </HStack>
+            </Pressable>
```
jason-gerard, line 222: Why is this duplicated

packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -180,20 +188,82 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
...
                 ) &&
                     (!isFilterSaved ||
                         !(
                             filterByTransaction === route.params.filters[0] &&
                             filterByDate === route.params.filters[1] &&
-                            filterByContact.length === 0
-                        )) && (
+                            filterByContact.length === 0 &&
+                        filterByTag.length === 0&&
+                        filterByTag.length === 0)) && (
```
jason-gerard, line 266: and this also?

packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -264,4 +335,14 @@ const styles = StyleSheet.create({
         marginLeft: "auto",
         marginRight: 5,
     },
+    tags: {
+        marginBottom: -13,
+    },
```
jason-gerard, line 340: What is this for?

jason-gerard (commented):
Good start, seems pretty messy. Confused by the duplicate code... Any reason for this?

jason-gerard:
You can run the linter with the command `yarn run lint:fix` in the root of the project

packages/client/screens/filters/FilterScreen.tsx
alexbara2000: good question, I think it is the rebase

jason-gerard (approved):

Andre-Ibrahim (approved):
To micro optimize the cli 's runtime you can do `yarn lint:fix` instead of `yarn run lint:fix`. This reduces the amount of characters the CLI needs to parse.

alexbara2000:
Thank you for the input. Further micro optimization can also be done if you run "yarn" instead of "yarn install". But that brings us to quantum computer science so it gets tricky.

PR #214, opened 2023-03-17
CRYP-269: Add Bitcoin wallet by scanning a QR code

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-269

## Changes

- Add a QR code icon in the wallet address input field in the `Add a Bitcoin Wallet` screen
- Ask the user the grant camera permission for the associated device when pressing the QR Code icon if not previously granted
    - Note: `"Expo Go" Would Like to Access the Camera` and the associated description will be replaced to the correct text as shown in the UI mockups when the app is built 
- Add a barcode scanner camera screen to allow the user to scan a Bitcoin wallet's QR code
- Add a `CurrencyTypeUILabels` enum that can be used to display the correct text format for the currency type names in the UI
- Add tests to make sure everything is working and matches UI mockups
- Add a `signIn()` test utils method (We should refactor all tests to use this method since all of them are copy and pasting the same code)

334137453_172450775612088_1870323913219142942_n
334592372_159628150020016_7593541348376667229_n
333374581_1227702031199750_5427358499279210083_n
333426722_226687619771057_6942686998771329474_n
333763653_1248450815878285_5622672659675623803_n
334000754_724139606106489_2791955619930165346_n


packages/client/components/QRCodeScannerInputIcon.tsx
```diff
@@ -0,0 +1,72 @@
...
+            } else {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                if (previousPermission.status == "undetermined" && !permission.granted) {
+                    return;
+                } else {
+                    cameraAccessDeniedAlert();
+                }
+            }
+        });
```
jason-gerard, line 57: Generally try to avoid callbacks especially if its scoped as async

```suggestion
        const permission = await requestPermission();
        if (permission.granted) {
            props.navigation.navigate("QRCodeScannerScreen", {
                fieldKey: props.fieldKey,
                setFieldValue: props.setFieldValue,
                currencyType: props.currencyType,
            });
        } else {
            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
            // @ts-ignore
            if (previousPermission.status == "undetermined" && !permission.granted) {
                return;
            } else {
                cameraAccessDeniedAlert();
            }
        }
```

packages/client/components/QRCodeScannerInputIcon.tsx
```diff
@@ -0,0 +1,72 @@
...
+        requestPermission().then((permission) => {
+            if (permission.granted) {
+                props.navigation.navigate("QRCodeScannerScreen", {
+                    fieldKey: props.fieldKey,
+                    setFieldValue: props.setFieldValue,
+                    currencyType: props.currencyType,
+                });
+            } else {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
```
jason-gerard, line 50: SonarQube will yell at is for using ts-ignore since it messes with the compiler is there an es-lint rule we can use directly instead? Maybe fix the issue if its easy enough too?

packages/client/screens/QRCodeScannerScreen.tsx
```diff
@@ -0,0 +1,241 @@
...
+                        </Box>
+                    );
+                },
+            });
+        }
+    };
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const handleBarCodeScanned = async ({ data, bounds }) => {
```
jason-gerard, line 59: I'm guessing the error was no implicit any, this should fix it

```suggestion
    const handleBarCodeScanned = async ({ data, bounds }: any) => {
```

packages/common/src/domain/currency_type_ui_labels.ts
```diff
@@ -0,0 +1,6 @@
+export enum CurrencyTypeUILabels {
+    ETHEREUM = "Ethereum",
+    BITCOIN = "Bitcoin",
+}
+
+export const currencyTypeUILabels = [CurrencyTypeUILabels.BITCOIN, CurrencyTypeUILabels.ETHEREUM] as const;
```
jason-gerard, line 6: Nit: Screaming snake case for constants

```suggestion
export const CURRENCY_TYPE_UI_LABELS = [CurrencyTypeUILabels.BITCOIN, CurrencyTypeUILabels.ETHEREUM] as const;
```

jason-gerard (approved):

packages/common/src/domain/currency_type_ui_labels.ts
domsec: We might have the same syntax in other similar files. Definitely something we should look at later this sprint since it'll be a small refactor.

narroarrow (approved):

PR #220, opened 2023-03-20
CRYP-200: As a User, I want to edit an Ethereum wallet, so that I can change the wallet name

alexbara2000:
-Added frontend for wallet name edit
-added backend for eth

image

image

image

### DIFF
--- a/packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
+++ b/packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
@@ -25,7 +25,7 @@ export class BtcEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
 
     async updateWallet(req: UpdateWalletRequest): Promise<WalletWithBalance> {
         const path = `users/${req.userId}/wallets/${req.address}`;
-        return this.request<WalletWithBalance>(Method.PUT, {}, path, null);
+        return this.request<WalletWithBalance>(Method.PUT, {}, path, req);
     }
 
     async getWallets(req: GetWalletsRequest): Promise<WalletWithBalance[]> {
--- a/packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
+++ b/packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
@@ -8,6 +8,7 @@ import { GetTransactionsRequest } from "@cryptify/common/src/requests/get_transa
 import { Transaction } from "@cryptify/common/src/domain/entities/transaction";
 import { Method } from "@cryptify/common/src/utils/gateway/abstract_gateway";
 import { DeleteWalletRequest } from "@cryptify/common/src/requests/delete_wallet_request";
+import { UpdateWalletRequest } from "@cryptify/common/src/requests/update_wallet_request";
 
 @Injectable()
 export class BtcEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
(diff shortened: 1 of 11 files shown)


packages/api/src/gateways/edge-gateway/eth_edge_gateway_strategy.ts
```diff
@@ -22,6 +23,11 @@ export class EthEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
         return this.request<WalletWithBalance>(Method.POST, {}, path, req);
     }
 
+    async updateWallet(req: UpdateWalletRequest): Promise<WalletWithBalance> {
+        const path = `users/${req.userId}/wallets/${req.address}`;
+        return this.request<WalletWithBalance>(Method.PUT, {}, path, req);
+    }
```
jason-gerard, line 29: You aren't passing any request here, this won't work

jason-gerard (commented):
Please test your code before submitting it for review.

packages/api/src/gateways/edge-gateway/eth_edge_gateway_strategy.ts
alexbara2000: It works

alexbara2000: I think you are referring to the the implementation for btc_edge that has null in the request which will not work. Btc implementation is another feature so I will do it in the other pull request..

packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
```diff
@@ -22,6 +23,11 @@ export class BtcEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
         return this.request<WalletWithBalance>(Method.POST, {}, path, req);
     }
 
+    async updateWallet(req: UpdateWalletRequest): Promise<WalletWithBalance> {
+        const path = `users/${req.userId}/wallets/${req.address}`;
+        return this.request<WalletWithBalance>(Method.PUT, {}, path, null);
```
jason-gerard, line 28: ?

packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
```diff
@@ -22,6 +23,11 @@ export class BtcEdgeGatewayStrategy extends AbstractEdgeGatewayStrategy {
...
+    }
```
Andre-Ibrahim, line 29: Update null to req before merging so you dont forget in your next pr

packages/eth-edge/src/services/wallets.service.ts
```diff
@@ -98,4 +99,17 @@ export class WalletsService {
 
         return new WalletBuilder().setWallet(wallet).setBalance(balance).build();
     }
+
+    async update(updateWalletReq: UpdateWalletRequest): Promise<WalletWithBalance> {
+        const [, balance] = await Promise.all([
+            this.walletRepository.update(
```
jason-gerard, line 105: This is fine but its better to use save() instead for the active object design pattern

jason-gerard (approved):

Andre-Ibrahim (approved):
Looks good just dont forget to fix the btc gateway before merging

packages/api/src/gateways/edge-gateway/btc_edge_gateway_strategy.ts
alexbara2000: To be done in the next PR

alexbara2000: nvm I did it

packages/eth-edge/src/services/wallets.service.ts
alexbara2000: You are right, using save instead of update is a well needed micro optimization

PR #221, opened 2023-03-20
CRYP-200: As a User, I want to edit a Bitcoin wallet, so that I can change the wallet name

alexbara2000:
Same as CRYP-201


jason-gerard (approved):
Good job. Can you fill out the description with the Jira ticket and changes

PR #249, opened 2023-03-31
CRYP-352: Fix transaction filter screens to matching UI mockups

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-352

## Fix

- Refactored `Filter by Transaction` to match UI mockups
- Moved `Filter by Date` to a separate screen
- Fixed saved filters not storing contacts and tags in the database
- Fixed the `Reset` button in the filter screen not properly resetting all filters across screens
- Fixed filter states not properly persisting between transaction list and filter screen(s)
- Updated tag filter badges in transaction list to match UI mockups
- Fix `Save this filter` not properly displaying/hiding 
- Refactor tag badge order to match specs
- Refactoring spacing/styling to match UI mockups
- Fixed all associated system tests

337008825_5525204817581441_6127166996984588362_n
337146897_618328433003381_3662064867418442761_n
337106307_1289887598324883_6885027930020036483_n
337298563_1211131732713176_5488726456627263067_n
337037127_231712736186305_6752166480453125614_n
334687952_180184301482257_7535258643205648768_n
337172777_2184726268390259_2390752876109899885_n

### DIFF
--- a/packages/client/screens/filters/FilterScreen.tsx
+++ b/packages/client/screens/filters/FilterScreen.tsx
@@ -51,7 +51,6 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
                 onPress={() => {
                     setFilterByTransaction(filtersByTransaction[0]);
                     setFilterByDate(filtersByDate[0]);
-                    route.params.setIsUsingSavedFilter(false);
                     setFilterByContact([]);
                     setFilterByTag([]);
                     setIsFilterSaved(false);
@@ -78,10 +77,10 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
                             onPress={() =>
                                 navigation.navigate("SavedFiltersScreen", {
                                     currencyType: route.params.wallet.currencyType,
-                                    setFilters: route.params.setFilters,
                                     setFilterByTransaction,
                                     setFilterByDate,
-                                    setIsUsingSavedFilter: route.params.setIsUsingSavedFilter,
+                                    setFilterByContact,
+                                    setFilterByTag,
                                     setIsFilterSaved,
                                 })
                             }
@@ -153,6 +152,7 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
                             setFilters: route.params.setFilters,
                             filterByDate,
                             setFilterByDate,
+                            setIsFilterSaved,
                         })
                     }
                 >
@@ -175,6 +175,7 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
                             setFilters: route.params.setFilters,
                             filterByContact,
                             setFilterByContact,
+                            setIsFilterSaved,
                         })
                     }
                 >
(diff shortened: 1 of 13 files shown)


narroarrow (approved):

packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -104,184 +92,196 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
...
-                    <Box style={styles.RadioItem} key={option}>
-                        <Radio key={option} value={option} color={"darkBlue.500"}>
-                            {option}
-                        </Radio>
-                    </Box>
-                ))}
-            </Radio.Group>
-        );
-    }
+    function renderFilterByContact() {
```
jason-gerard, line 95: you can just do filterByContact.join(", ") here

packages/client/screens/filters/FilterScreen.tsx
```diff
@@ -104,184 +92,196 @@ export default function FilterScreen({ route, navigation }: HomeStackScreenProps
...
+            </ScrollView>
+            <Box style={styles.applyFiltersButtonContainer}>
                 <Button
-                    style={styles.applyButton}
                     isDisabled={
                         filterByTransaction === route.params.filters[0] &&
                         filterByDate === route.params.filters[1] &&
-                        filterByContact.length === 0 &&
-                        filterByTag.length === 0
+                        JSON.stringify(filterByContact) === JSON.stringify(route.params.contactFilters) &&
```
jason-gerard, line 259: Why change checking length to the values?

jason-gerard (approved):
Looks good, since some of the filtering was changed specifically the custom filter option. does that effect the saved filter feature or the data passed is still the same? They are pretty couples unfortunately so there might be some cascading going on

packages/client/screens/filters/FilterScreen.tsx
domsec: Because of the way the contact and tag filters were originally created, there were scenario's where even if the `filterByContact`/`filterByTag` were "empty", we'd still want the button to be enabled since those changes no longer persist to the transaction list until pressing the `Apply filters` button. So for example if you removed all such filters, you wouldn't be able to apply those changes to the transaction list and thus the transaction list would still have those old filters applied. 

Hence why this check is needed.

domsec:
> Looks good, since some of the filtering was changed specifically the custom filter option. does that effect the saved filter feature or the data passed is still the same? They are pretty couples unfortunately so there might be some cascading going on

I made sure to not mess with it at all because like you said they are all very coupled together. So there's extra pre-processing going on unfortunately to make sure everything remains intact. The only change done to the save filter stuff was making sure the contacts and tags are actually getting saved in the db whereas they never were before.

packages/client/screens/filters/FilterScreen.tsx
domsec: Good catch, for some reason I didn't think of this.

PR #250, opened 2023-03-31
CRYP-271: Remove ungranted QR code camera permissions system tests

domsec:
JIRA: https://cryptify.atlassian.net/browse/CRYP-271

## Changes

- All QR code ungranted camera permissions have been removed as outlined in the added comments to each system test file.
   - In short, these tests cannot be executed due to the limitation of Detox in relation to Android emulators (camera permissions are always granted and cannot be changed) and our current inability to run such tests on an iOS emulator. Therefore, having a test that is always guaranteed to pass based on the old code would be useless.


jason-gerard (approved):

jason-gerard (commented):
Lint fix

Andre-Ibrahim (approved):
makes sense 👍

PR #251, opened 2023-03-31
CRYP-000: Add gif recordings for release 3

jason-gerard:


domsec (approved):
Looks good. Just confirm that the removed QR code related system tests are included. I don't see them but just want to make sure that is indeed the case.

jason-gerard:
> Looks good. Just confirm that the removed QR code related system tests are included. I don't see them but just want to make sure that is indeed the case.

Yea since they were skipped on Android they won't show either way

Andre-Ibrahim (approved):

